PR #129, opened 2019-10-11
Add proxy sharing function. Fix potential security vulnerbility.

galvasis193:
The pull request has two commits.

commit 1.Add proxy sharing function and block unauthorized access vulnerability.
commit 2.When settings change proxy used to be relying manual restart in order for the change to be applied. Now in this commit, v2rayNG does this automatically.


galvasis193:
@2dust

PR #320, opened 2020-03-21
Re-arrange process in order to save memory

yuhan6665:
- V2Ray core service is moved to main process
- All activities are moved to UI process
- Tile service stay at main process
- Also, adjust white space to match with Android Studio standards

Before:
before1
After:
after

v2rayNG has a good design with multiple processes. However, I found even if user wan't look at the v2rayNG, as soon as user expand status bar to see tiles, system will launch main process. As you can see on my device both processes take at least 300MB memory.
I tried to move TileService to a different process but that didn't help.
I think we should inverse it, assign activities to a separate process. User can still activate vpn from the tile. During my test, the UI process is never launched unless if you navigate to the app. This could cut memory usage by almost half.

### DIFF
--- a/V2rayNG/app/src/main/AndroidManifest.xml
+++ b/V2rayNG/app/src/main/AndroidManifest.xml
@@ -22,16 +22,17 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <!-- <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" /> -->
     <application
-        android:name=".AngApplication"
-        android:allowBackup="true"
-        android:icon="@mipmap/ic_launcher"
-        android:label="@string/app_name"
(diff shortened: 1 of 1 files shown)


2dust:
Disagree with you
When multiple process applications are terminated by the system low memory killer, you can try to make a certain process of the application survive
多个进程应用在被系统low memory killer终止时，可以尽量让应用的某个进程能存活

yuhan6665:
Thanks for your review.
This is what I observe, if you try simulate the system low memory by killing "com.v2ray.ang" process. System often relaunch it, even if you don't open v2rayNG. This process is of no use except for TileService. Isn't it inefficient to use that much memory to display a button?
On the other hand, maybe the big memory usage is due to the Application object is heavy. Are you open to improvements for that?

PR #345, opened 2020-04-03
Sort test results

yuhan6665:
Add a nice animation for sorting. I've tested sort > 40 nodes.
I know you have another "test real delay" function. I've made the methods, it should be straight forward for you to integrate with :)


2dust:
Thank you for your pr
Sorting function requires global redesign

yuhan6665:
Ok, let me know if you have plan to do it

PR #363, opened 2020-04-11
All inbounds speed

yuhan6665:
The first commit
- Parse inbound tags for preload Json and custom config
- When more than two tags available, show the one with higher traffic
- Also, improve the notification title so that the text won't bounce too much

The second
- When user save custom config and don't have a tag, show a toast for warning
- Also, improve the error message toast when custom config is malformed 

I've tested the notification speed works for socks and http inbound


2dust:
Thank you for your pr
This function is too complicated for  users

PR #380, opened 2020-04-19
Add tls certificate settings for vmess

yuhan6665:
~~Configuration is added according to 
https://android-developers.googleblog.com/2016/07/changes-to-trusted-certificate.html
I was only able to verify the system log  and see the config was in effect. I can't find a testing environment.~~

### DIFF
--- a/V2rayNG/app/build.gradle
+++ b/V2rayNG/app/build.gradle
@@ -18,6 +18,7 @@ android {
         multiDexEnabled true
         versionCode 212
         versionName "1.0.2"
+        vectorDrawables.useSupportLibrary true
     }
 
     signingConfigs {
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/dto/AngConfig.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/dto/AngConfig.kt
@@ -17,6 +17,7 @@ data class AngConfig(
                          var requestHost: String = "",
                          var path: String = "",
                          var streamSecurity: String = "",
+                         var tlsCertificateFile: String = "",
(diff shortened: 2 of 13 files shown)


yuhan6665:
~~Tested the fix is not working for self-signed cert. Looks like v2ray core will not load user cert by default. Still working on it.~~

yuhan6665:
In recent release, the feature is added to allow user to choose verifying cert. However, this feature is not working for people using self signed cert or, Let's Encrypt cert is not recognised by their version of OS.
I did some tests. Looks like Golang x509.SystemCertPool() doesn't fetch user cert. So the only option is to add the cert File to the Json config.
I added a File chooser for this. Note since the File chooser will not return the actual file path, we need to copy it to app folder.

Also, update the vector icon reference to use "app:srcCompat".

yuhan6665:
Closed as discussed with @2dust

robertlee22:
v2rayNG(v1.47/Core v4.30.0) doesn't see the the "tls_cert" button when add a server.  Did this pull request be merged in previous release? I checked all releases and didn't find tags about #380 pull request.

yuhan6665:
@primelyw closed means not merged.

robertlee22:
@yuhan6665 Thanks for your reply

robertlee22:
@yuhan6665 Do you have any idea about how to let app trust my own CA certificate which has been installed on my android phone?

yuhan6665:
@primelyw if you know how to use custom config you can embed the full content of certificate in it. https://www.v2fly.org/config/transport.html#tlsobject

robertlee22:
Config perfectly, Thanks again! @yuhan6665

PR #516, opened 2020-07-22
add support to new version vmess link

iseki0:
some information can be found at: https://github.com/v2ray/discussion/issues/720

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
@@ -2,7 +2,6 @@ package com.v2ray.ang.util
 
 import android.graphics.Bitmap
 import android.text.TextUtils
-import android.util.Log
 import com.google.gson.Gson
 import com.v2ray.ang.AngApplication
 import com.v2ray.ang.AppConfig
@@ -16,12 +15,9 @@ import com.v2ray.ang.AppConfig.VMESS_PROTOCOL
 import com.v2ray.ang.R
 import com.v2ray.ang.dto.AngConfig
 import com.v2ray.ang.dto.VmessQRCode
-import com.v2ray.ang.extension.defaultDPreference
-import org.jetbrains.anko.toast
+import java.net.URI
 import java.net.URLDecoder
 import java.util.*
-import java.net.*
-import java.math.BigInteger
 
 object AngConfigManager {
     private lateinit var app: AngApplication
@@ -256,7 +252,11 @@ object AngConfigManager {
             if (server.startsWith(VMESS_PROTOCOL)) {
 
                 val indexSplit = server.indexOf("?")
-                if (indexSplit > 0) {
+                val newVmess = tryParseNewVmess(server)
+                if (newVmess != null) {
+                    vmess = newVmess
+                    vmess.subid = subid
+                } else if (indexSplit > 0) {
                     vmess = ResolveVmess4Kitsunebi(server)
                 } else {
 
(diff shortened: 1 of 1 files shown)


V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
```diff
@@ -378,6 +378,60 @@ object AngConfigManager {
...
+                            .matchEntire(uri.userInfo)?.groupValues
+                            ?: error("parse user info fail.")
+            val tls = tlsStr.isNotBlank()
+            val queryParam = uri.rawQuery.split("&")
+                    .map { it.split("=").let { (k, v) -> k to URLDecoder.decode(v, "utf-8")!! } }
+                    .toMap()
+            val vmess = AngConfig.VmessBean()
+            vmess.address = uri.host
+            vmess.port = uri.port
+            vmess.guid = uuid
```
yuhan6665: guid is internal, set vmess.id instead

V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
```diff
@@ -378,6 +378,60 @@ object AngConfigManager {
...
+
+            when (protocol) {
+                "tcp" -> {
+                    vmess.network = "tcp"
+                    vmess.headerType = queryParam["type"] ?: "none"
+                    vmess.requestHost = queryParam["host"] ?: ""
+                }
+                "http" -> {
+                    vmess.network = "h2"
+                    vmess.path = queryParam["path"]?.takeIf { it.trim() != "/" } ?: ""
```
yuhan6665, line 412: I understand "/" is a valid path

V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
```diff
@@ -378,6 +378,60 @@ object AngConfigManager {
...
+                    vmess.requestHost = queryParam["host"]?.split("|")?.get(0) ?: ""
+                }
+                "kcp" -> {
+                    vmess.network = "kcp"
+                    vmess.headerType = queryParam["type"] ?: "none"
+                    vmess.path = queryParam["seed"] ?: ""
+                }
+                "quic" -> {
+                    vmess.network = "quic"
+                    vmess.security = queryParam["security"] ?: "none"
```
yuhan6665: Currently quic security should be put as vmess.requestHost.
This brings a question.. how does the new format carry vmess.securiity?

iseki0: 现阶段，vmess的security取消，默认auto

V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
iseki0: 因为他们使用 / 作为默认值，我担心传递 / 会造成麻烦

V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
yuhan6665: Is the new format finalised? If not I'd suggest we should add security.

iseki0: Until now, looks there is no plan for it.  And I don't think we should add this feature. Because it can auto negotiation.

ghost: no matter which encryption has been choosed, client can always connect to server, so I don't think security option is necessary.

ghost: Hello.
As far as I could tell, VMess could perform auto-negotiation and they suggest setting it to `auto`. So this is not required anymore.

image

yuhan6665: Ok, it is not mandatory but it would be a convenient setting. "none" is a common choice with tls. Also user would expect to transfer basic settings with export-and-then-import. For instance..
https://github.com/2dust/v2rayNG/issues/356
dalao @rprx was surprised by it :)

ghost: That only happens when using the old standard. It's your parser's fault not to work correctly.
The new standard obsoletes this setting, so they are irrelevant.

iseki0: The new version scheme has removed the paramter. I think it's viable to user, because of we don't use base64. So that user will understand what's happeded.

Issue #536, opened 2020-08-05
1.2.16 not work after sleep?


mactrash:
在提出问题前请先自行排除服务器端问题，同时也请通过搜索确认是否有人提出过相同问题。

### 预期行为
It should be keep connection as long as I disconnect

### 实际行为
The connection broken after sometime after sleep... sometime it do work normally, need restart the app to reconnect

mactrash:
Anyone meet the same after upgrade? of just me?

yuhan6665:
Can you post the adb log when you experience connection broken?

Loukky:
同样的问题，不过刚刚推送了一个新版本，看更新日志是bug fix，不知道有没有修复这个问题

yuhan6665:
@Loukky yes it is fixed in 1.2.17

Issue #532, opened 2020-08-02
新版本经常断连


brunofeng:
最近版本发生一些怪事，之前是没有的，
状况1：比如从一个能翻墙的网络，到一个不能翻墙的网络，再回到能翻墙的网络，就必须手动断开再连上才行，以前就不用

状况2：在一个能翻墙的网络里，过了很久之后，会莫名其妙断掉，需要断开再连

### 日志信息
<details>
--------- beginning of main
08-02 21:49:36.870 I/GoLog   (27197): PrepareDomain err: lookup fpg.csteam.cc: No address associated with hostname
08-02 21:49:36.870 I/GoLog   (27197): PrepareDomain err: lookup fpg.csteam.cc: No address associated with hostname
08-02 21:49:36.870 I/GoLog   (27197): PrepareDomain err: lookup fpg.csteam.cc: No address associated with hostname
08-02 21:49:36.871 I/GoLog   (27197): PrepareDomain err: lookup fpg.csteam.cc: No address associated with hostname
08-02 21:49:36.871 I/GoLog   (27197): PrepareDomain err: lookup fpg.csteam.cc: No address associated with hostname
08-02 21:49:36.871 I/GoLog   (27197): PrepareDomain err: lookup fpg.csteam.cc: No address associated with hostname
08-02 21:49:36.871 I/GoLog   (27197): PrepareDomain err: lookup fpg.csteam.cc: No address associated with hostname
08-02 21:49:37.089 I/GoLog   (27197): Preparing Domain: fpg.csteam.cc:21556
08-02 21:49:38.881 I/GoLog   (27197): PrepareDomain maxRetry reached. exiting.
08-02 21:49:38.881 I/GoLog   (27197): PrepareDomain maxRetry reached. exiting.
08-02 21:49:39.319 I/GoLog   (27197): Using Prepared: 104.238.136.133
08-02 21:49:39.320 I/GoLog   (27197): Using Prepared: 104.238.136.133
08-02 21:49:40.096 I/GoLog   (27197): Preparing Domain: fpg.csteam.cc:21556
08-02 21:49:41.097 I/GoLog   (27197): Preparing Domain: fpg.csteam.cc:21556
08-02 21:49:46.099 I/GoLog   (27197): Preparing Domain: fpg.csteam.cc:21556
08-02 21:49:47.087 I/GoLog   (27197): PrepareDomain err: i/o timeout
[...]

PR #537, opened 2020-08-06
Revert "Refresh prepared domain every 30 minutes"

yuhan6665:
This reverts commit 903352ec9c4231131ea2ad75b68661526a47d156.

### DIFF
--- a/AndroidLibV2rayLite/VPN/vpnservice_support.go
+++ b/AndroidLibV2rayLite/VPN/vpnservice_support.go
@@ -24,7 +24,6 @@ type resolved struct {
 	domain       string
 	IPs          []net.IP
 	Port         int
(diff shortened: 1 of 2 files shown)


IceCodeNew:
> This reverts commit [903352e](https://github.com/2dust/v2rayNG/commit/903352ec9c4231131ea2ad75b68661526a47d156).

Why?

yuhan6665:
@IceCodeNew in #474 I added a DNS refresher but it has a bug. It caused #532 #536.
Technically it is possible to do a fix, for example https://github.com/2dust/AndroidLibV2rayLite/pull/16
However after discussion with @2dust, we think it is a complex feature that may cause other problems meanwhile most people won't change their ip address often. You just need to restart service in that case.

PR #598, opened 2020-09-12
Add missing process for ScSwitchActivity

yuhan6665:
In a recent change, I've simplified ScSwitchActivity but forgot to move it to daemon process.


fff111222:
Please add subscription remarks id to server list app menu

PR #387, opened 2020-04-24
Target sdk 29

yuhan6665:
We will need to update target sdk to latest Android version this year. The main issue with the current code is that tun2socks binary will not be able to run from app folder. This is stated in:
https://issuetracker.google.com/issues/128554619#comment4

In this pull request, I've moved the tun2socks so that they can be copied to where .so lib is. A change is made to Go project with the new path. Note I've made minimum code change for this fix. You can probably substitute the aar.
If you are ok with it. I'm going to do more commits to clean up. Including the duplicated copy of tun2socks in app folder and its logic.

PR #626, opened 2020-09-28
Clean up tun2socks

yuhan6665:
Since tun2socks has been moved to jniLibs folder in https://github.com/2dust/v2rayNG/pull/387, we don't need binaries in AndroidLibV2rayLite project any more. The related logic can also be removed.


rurirei:
you should use applicationContext.nativeDir https://github.com/2dust/v2rayNG/blob/df3f1ca3efdd5b6e9115cbe7ecdb4260541ab732/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/Utils.kt#L416

yuhan6665:
@rurirei you might be confused by the method name. So currently Golang need two paths from Kotlin side. One is the tun2socks executable path, this is what you mentioned. The other is a local file for the vpn interface file descriptor. This file is in our sandbox.

rurirei:
@yuhan6665 sorry, mistake of my memories..

Issue #609, opened 2020-09-19
安卓10的无限retry修复


DametLiu:
1.在VpnService类中，高版本的kotlin有一个mInterface没有被初始化的异常，必须初始化为可空变量?
2.在setup函数中，以下代码
`try {
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
                connectivity.requestNetwork(defaultNetworkRequest, defaultNetworkCallback)
                listeningForDefaultNetwork = true
            }
        } catch(e:Exception) {}`
必须用try catch包裹。因为这段代码会导致后续代码无法执行，也没有抛异常（顶层抓住了）。
这段代码就算是有问题，也不影响后续代码，否则得话，sendFd就会无限retry。这个逻辑应该可以优化。

DametLiu:
我猜测他们报告得一部分vpn可以连接但没有网络得问题多半是因为这个，我在本地查看localsocket得文件，因为文件描述符被close，后续得establish没有赋值导致sendFd出现了bad 异常。所以文件是没有数据写入得。

yuhan6665:
@DametLiu looks like I can't reproduce this issue. Can you post adb log?
You seems to have a good understanding of the problem. You are welcome to push a fix also :)

DametLiu:
> @DametLiu looks like I can't reproduce this issue. Can you post adb log?
> You seems to have a good understanding of the problem. You are welcome to push a fix also :)

u can replace codes in V2RayVpnService.setup, i havent this project permission 

        // fix sometimes this code will throw an exception, then vpn cannot work correctly
        // the exception must catch here, else will get a bad file descriptor,
        // because after this code thrown, mInterface cannot get a new value, it's null
        // we'll get two kinds of logs
        // 1. Bad descriptor then sendFd retry 1234512345
        // 2. lateinit var mInterface not initialized and retry 1234512345
        try {
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
                connectivity.requestNetwork(defaultNetworkRequest, defaultNetworkCallback)
                listeningForDefaultNetwork = true
            }
        } catch (e: Exception) {}

yuhan6665:
@DametLiu, thanks for your suggestion. Though, I'm more interested in what is the original exception throw by requestNetwork(). Try catch is ok but we should understand the root cause.

PR #635, opened 2020-10-01
issues of VPNService.kt

rurirei:
https://github.com/2dust/v2rayNG/issues/609

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt
@@ -58,8 +58,6 @@ class V2RayVpnService : VpnService(), ServiceControl {
         }
     }
 
-    private var listeningForDefaultNetwork = false
-
     override fun onCreate() {
         super.onCreate()
 
@@ -155,7 +153,6 @@ class V2RayVpnService : VpnService(), ServiceControl {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
             try {
                 connectivity.requestNetwork(defaultNetworkRequest, defaultNetworkCallback)
-                listeningForDefaultNetwork = true
             } catch (ignored: Exception) {
                 // ignored
             }
@@ -207,7 +204,6 @@ class V2RayVpnService : VpnService(), ServiceControl {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
             try {
                 connectivity.unregisterNetworkCallback(defaultNetworkCallback)
-                listeningForDefaultNetwork = false
             } catch (ignored: Exception) {
                 // ignored
             }
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt
@@ -163,7 +163,13 @@ class V2RayVpnService : VpnService(), ServiceControl {
         }
 
         // Create a new interface using the builder and save the parameters.
-        mInterface = builder.establish()
+        try {
+            mInterface = builder.establish()!!
+        } catch (e: Exception) {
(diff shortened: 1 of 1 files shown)


V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt
```diff
@@ -200,9 +200,12 @@ class V2RayVpnService : VpnService(), ServiceControl {
 //        val emptyInfo = VpnNetworkInfo()
 //        val info = loadVpnNetworkInfo(configName, emptyInfo)!! + (lastNetworkInfo ?: emptyInfo)
 //        saveVpnNetworkInfo(configName, info)
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && listeningForDefaultNetwork) {
-            connectivity.unregisterNetworkCallback(defaultNetworkCallback)
-            listeningForDefaultNetwork = false
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            try {
+                connectivity.unregisterNetworkCallback(defaultNetworkCallback)
```
yuhan6665, line 211: I feel like we can simply check null for defaultNetworkCallback then we don't need try catch?

V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt
```diff
@@ -153,8 +151,11 @@ class V2RayVpnService : VpnService(), ServiceControl {
         }
 
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-            connectivity.requestNetwork(defaultNetworkRequest, defaultNetworkCallback)
-            listeningForDefaultNetwork = true
+            try {
+                connectivity.requestNetwork(defaultNetworkRequest, defaultNetworkCallback)
+            } catch (ignored: Exception) {
+                // ignored
```
yuhan6665, line 157: Let's not ignore this exception by logging it. By the way, if you know the exact exception, we should see if there are other ways to fix the root problem.

yuhan6665 (commented):
Thanks for your pull request! left two small comments.

V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt
rurirei: catch is needed for avoiding crash. the reason of IOException is something i do not care (actually Android has too much addition like this). exception at registerDefaultNetwork could be simply ignored even now.

yuhan6665: Fine, we don't care for unregister(). But we should care for register(). We have this code for a reason and some users are depending on it. We should at least add a log for exception. See more details
https://github.com/shadowsocks/shadowsocks-android/commit/3a8f863bd1f0deed081de082a674020368f30319

rurirei: mostly `try` is needed, catch IOException and analysis is maybe a way.

rurirei (commented):
third commit:

call sendFd() before tun2socks started leads to java.io.IOException: connection refused at localSocket() ..
https://github.com/2dust/v2rayNG/blob/df3f1ca3efdd5b6e9115cbe7ecdb4260541ab732/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt#L178
even sendFd() called twice at https://github.com/2dust/v2rayNG/blob/df3f1ca3efdd5b6e9115cbe7ecdb4260541ab732/AndroidLibV2rayLite/interact.go#L240 https://github.com/2dust/v2rayNG/blob/df3f1ca3efdd5b6e9115cbe7ecdb4260541ab732/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt#L166
so it needs called only once after escort.run() hangs tun2socks up.

PR #662, opened 2020-10-09
Add external geofile replacement help

yuhan6665:
https://github.com/yuhan6665/v2rayNG/tree/ext-geo

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,16 @@ A V2Ray client for Android
 
 ### Usage
 
-See our [wiki](https://github.com/2dust/v2rayNG/wiki)
+#### Geoip and Geosite
+v2rayNG release already embedded domain file `geoip.dat` and `geosite.dat`. However it is (probably) not the latest and not the most complete list.
+For power user, the embedded files can be easily replaced with the following steps:
+1. Launch v2rayNG (v1.4.9+)
+2. Find existing geoip.dat and geosite.dat in `Android/data/com.v2ray.ang/files/assets` (path may differ on some Android device)
+3. Replace them with the latest [domain list](https://github.com/v2fly/domain-list-community) and [ip list](https://github.com/v2fly/geoip)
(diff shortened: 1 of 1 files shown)


rurirei:
you need to use SAF to provide files, as external app dir is restricted to access on api 30

yuhan6665:
> you need to use SAF to provide files, as external app dir is restricted to access on api 30

You are taking about deprecation of https://developer.android.com/reference/android/os/Environment#getExternalStoragePublicDirectory(java.lang.String)?
The proposal is to use https://developer.android.com/reference/android/content/Context#getExternalFilesDir(java.lang.String)

rurirei:
@yuhan6665 no, study on new storage on Android 11please.

yuhan6665:
> @yuhan6665 no, study on new storage on Android 11please.

@rurirei thanks for your information. I read the [document](https://developer.android.com/training/data-storage) and did some test on Android 11 emulator. So v2rayNG can still access `Android/data/com.v2ray.ang/` but other apps (including system file viewer) can't. That means user will not be able to update this file.

Looks like we have no other choice but to add a File chooser for getting user's geofiles. Fortunately we can keep building on top of #662. I think it's best to add another item in advanced settings. Thoughts? @2dust

ayanamist:
按 #172 的情况，这里的代码已经不是最新了，这个PR可能没有什么意义……

pansila:
> > @yuhan6665 no, study on new storage on Android 11please.
> 
> @rurirei thanks for your information. I read the [document](https://developer.android.com/training/data-storage) and did some test on Android 11 emulator. So v2rayNG can still access `Android/data/com.v2ray.ang/` but other apps (including system file viewer) can't. That means user will not be able to update this file.
> 
> Looks like we have no other choice but to add a File chooser for getting user's geofiles. Fortunately we can keep building on top of #662. I think it's best to add another item in advanced settings. Thoughts? @2dust

Why adding a custom data updater hasn't been added in the latest V2rayNG(1.6.15)? Is there any further restriction from Android 11?

PR #670, opened 2020-10-14
Fix crash when custom config cannot be parsed by Gson

yuhan6665:


V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/V2rayConfigUtil.kt
```diff
@@ -64,11 +64,16 @@ object V2rayConfigUtil {
         if (TextUtils.isEmpty(jsonConfig)) {
             return null
         }
-        val v2rayConfig = Gson().fromJson(jsonConfig, V2rayConfig::class.java) ?: return null
+        val v2rayConfig: V2rayConfig = try {
```
rurirei:
```kotlin
        val v2rayConfig: V2rayConfig? = try {
            Gson().fromJson(jsonConfig, V2rayConfig::class.java)
        } catch (e: com.google.gson.JsonSyntaxException) {
            e.printStackTrace()
            null
        }
        for (outbound in v2rayConfig?.outbounds) {
            if (outbound.protocol.equals(EConfigType.VMESS.name, true) ||
                    outbound.protocol.equals(EConfigType.SHADOWSOCKS.name, true) ||
                    outbound.protocol.equals(EConfigType.SOCKS.name, true)) {
                return outbound
            }
        }
```

rurirei (changes requested):

V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/V2rayConfigUtil.kt
yuhan6665: Code looks better now. Thanks for your review

PR #672, opened 2020-10-15
useless multiDex

rurirei:
remove useless multiDex http://developer.android.com/tools/building/multidex.html
>> the feature is: it allows your complex app to compile. The scenarios for using it are when your app fails to compile due to hitting the 64K DEX method reference limit. This appears as a build error, such as: Conversion to Dalvik format failed: Unable to execute dex: method ID not in [0, 0xffff]: 65536

### DIFF
--- a/V2rayNG/app/build.gradle
+++ b/V2rayNG/app/build.gradle
@@ -15,7 +15,6 @@ android {
         applicationId "com.v2ray.ang"
         minSdkVersion 17
         targetSdkVersion Integer.parseInt("$targetSdkVer")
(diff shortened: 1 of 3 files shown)


yuhan6665:
@rurirei I added multiDex in order to support Android 4.3 and 4.4. We can only remove it If we drop support for older Android.

rurirei:
> @rurirei I added multiDex in order to support Android 4.3 and 4.4. We can only remove it If we drop support for older Android.

that's it.. you can wait until that support dropped

PR #689, opened 2020-10-18
Release AndroidLibV2rayLite 22

yuhan6665:
Fix the compile error after #657


yuhan6665:
will push new PR to include #695 so that we don't need to release yet another version

Issue #589, opened 2020-09-06
[Feature Request ] Bandwidth Consumption


chromer030:
Hi , It will be great and useful if V2rayNG calculate and show bandwidth Consumption like Shadowsocks android client for every server , so user will now how much traffic is consumed.

Shadowsocks example screenshot :

![Shadowsocks_v4 7 4](https://user-images.githubusercontent.com/13361552/92329326-004eda00-f07c-11ea-8761-f2986ad4a792.png)

PR #701, opened 2020-10-23
Traffic Stats base

rurirei:
https://github.com/2dust/v2rayNG/issues/692 https://github.com/2dust/v2rayNG/issues/589 this is the base & core implementation for traffic stats, then two leaving there: UI to show user how much traffic he uses, and your thought of if differ from proxyTraffic and directTraffic to two item.

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayServiceManager.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayServiceManager.kt
@@ -308,6 +308,7 @@ object V2RayServiceManager {
                         val queryTime = System.currentTimeMillis()
                         val sinceLastQueryInSeconds = (queryTime - lastQueryTime) / 1000.0
                         var proxyTotal = 0L
+                        var directTotal = 0L
                         val text = StringBuilder()
                         outboundTags.forEach {
                             val up = v2rayPoint.queryStats(it, "uplink")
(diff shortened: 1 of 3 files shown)


V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayServiceManager.kt
```diff
@@ -319,17 +321,19 @@ object V2RayServiceManager {
...
                                 appendSpeedString(text, outboundTags.firstOrNull(), 0.0, 0.0)
                             }
                             appendSpeedString(text, TAG_DIRECT, directUplink / sinceLastQueryInSeconds,
                                     directDownlink / sinceLastQueryInSeconds)
-                            updateNotification(text.toString(), proxyTotal, directDownlink + directUplink)
+                            updateNotification(text.toString(), proxyTotal, directTotal)
                         }
                         lastZeroSpeed = zeroSpeed
                         lastQueryTime = queryTime
+                        AngConfigManager.updateTrafficStats(proxyTotal + directTotal)
```
yuhan6665, line 336: Thanks for your pr. I wanted to add this feature for a while, but was hesitate to do so. For the following reason:
Currently v2rayNG stores the entire list of servers under a single key in the SharedPreference store. This is bad that shared preference is not designed for this. I suspect it is the root cause of #535. Obviously adding frequent update for this list is not a good idea.
I was thinking maybe we should do a data migration, to a storage with one key <-> one server. We should think carefully and consider other optimisation as well.
CC @2dust

rurirei: Preferences behavior needs a rewrite, the main causes is of angConfig & AngConfigManager #663

yuhan6665: We had some chats. We don't plan to add traffic stats at the moment. Though we are looking into possible improvements to server list storage. Including this https://android-developers.googleblog.com/2020/09/prefer-storing-data-with-jetpack.html

rurirei: the dataStore does not much than sharedPreferences, recommend to dataBase instead. https://developer.android.com/reference/android/database/package-summary

PR #725, opened 2020-11-02
fix: the problem that importing SS URL not support standard format

mzz2017:
For example, an address can be generated with path '/' and it will be like ss://username:password@example.com:port/#anchor

v2rayNG cannot import it while v2rayN can.

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
@@ -318,7 +318,7 @@ object AngConfigManager {
                     result = Utils.decode(result)
                 }
 
-                val legacyPattern = "^(.+?):(.*)@(.+?):(\\d+?)$".toRegex()
+                val legacyPattern = "^(.+?):(.*)@(.+?):(\\d+?)/?$".toRegex()
                 val match = legacyPattern.matchEntire(result)
                 if (match == null) {
                     return R.string.toast_incorrect_protocol


V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
```diff
@@ -318,7 +318,7 @@ object AngConfigManager {
                     result = Utils.decode(result)
                 }
 
-                val legacyPattern = "^(.+?):(.*)@(.+?):(\\d+?)$".toRegex()
+                val legacyPattern = "^(.+?):(.*)@(.+?):(\\d+?)/?$".toRegex()
```
yuhan6665, line 321: Thanks for your pr! Instead of `/?` I think adding `(\\/(.*)?)?` is better?
Since it also match an url with plugin parameter
`ss://cmM0LW1kNTpwYXNzd2Q=@192.168.100.1:8888/?plugin=obfs-local%3Bobfs%3Dhttp#Example2`

mzz2017: Thanks for review!
But I have some questions.
Are plugins of shadowsocks in v2rayNG supported? I mean that if we support importing it, it will not work either.
And I guess that escape character `\` is not needed for `/`.
However, your proposal is harmless on the whole. Feel free to add your commits if you like to.

yuhan6665: yeah to get full support we may need to parse ss-v2ray plugin parameter and put as "streamSettings" in json config. It is more complex so your proposal is ok to me.
For the escape character, my online helper say it is needed:
https://regex101.com/r/uBYrMn/1/

mzz2017: If the regex is wrapped with `/`, it should escape, like PHP and ECMAScript.

yuhan6665: Tried in Android Studio, says it is redundant. Looks good. 👍

Issue #728, opened 2020-11-06
请问 android客户端使用vmess协议 节点信息配置正常 启动vpn服务并代理成功需要大概12秒


alijhen:
android客户端扫描二维码导入配置信息 点击按钮启动vpn服务

### 预期行为
希望会快一些 可能这里描述不清楚 下边有log日志 

### 实际行为
2020-11-07 02:30:32.721 20663-20710/com.v2ray.ang I/GoLog: Prepare Result:
     Domain: my domain
     Port: my port
     IPs: [172.67.162.249 104.24.116.236 104.24.117.236 2606:4700:3036::6818:74ec 2606:4700:3031::ac43:a2f9 2606:4700:3031::6818:75ec]
2020-11-07 02:30:42.664 20663-20707/com.v2ray.ang I/GoLog: EscortRun cmd.Wait err: exit status 1
2020-11-07 02:30:42.665 20663-20707/com.v2ray.ang I/GoLog: EscortRun Unexpected Exit, Restart now.
2020-11-07 02:30:42.668 20663-20707/com.v2ray.ang I/GoLog: EscortRun Waiting....
关键点在这里：GoLog: Prepare Result的打印时间为 2020-11-07 02:30:32 到最下边的日志打印时间差为10秒

### 复现方法
1.使用v2rayNG源码在android studio上导入项目并运行程序
2.扫描二维码导入配置信息
3.点击右下角连接按钮

### 日志信息
<details>
通过 `adb logcat -s com.v2ray.ang GoLog V2rayConfigUtilGoLog Main` 获取日志。请自行删减日志中可能出现的敏感信息。

如果问题可重现，建议先执行`adb logcat -c`清空系统日志再执行上述命令，再操作重现问题。
```
在这里粘贴日志
```2020-11-07 02:30:32.025 19840-19840/com.v2ray.ang D/V2rayConfigUtilGoLog: {详细信息没有复制到这里}
     
2020-11-07 02:30:32.643 20663-20710/com.v2ray.ang I/GoLog: Preparing Domain: w4euvsas.com:443
2020-11-07 02:30:32.643 20663-20663/com.v2ray.ang I/GoLog: file exists
2020-11-07 02:30:32.643 20663-20663/com.v2ray.ang I/GoLog: EnableLocalDNS: false
    ForwardIpv6: false
    DomainName: domain:port
2020-11-07 02:30:32.643 20663-20663/com.v2ray.ang I/GoLog: loading v2ray config
[...]

PR #732, opened 2020-11-08
isTun2socksRunning bool

rurirei:
more status check is needed, as #728

### DIFF
--- a/AndroidLibV2rayLite/CoreI/Status.go
+++ b/AndroidLibV2rayLite/CoreI/Status.go
@@ -6,6 +6,7 @@ import (
 
 type Status struct {
 	IsRunning       bool
+	IsTRunning      bool
 	PackageName     string
 	PackageCodePath string
 
--- a/AndroidLibV2rayLite/interact.go
+++ b/AndroidLibV2rayLite/interact.go
@@ -161,11 +161,13 @@ func (v *V2RayPoint) pointloop() error {
 	v.SupportSet.Setup(v.status.GetVPNSetupArg(v.EnableLocalDNS, v.ForwardIpv6))
 	v.SupportSet.OnEmitStatus(0, "Running")
 
+	isTRunning := false
 	if !v.ProxyOnly {
 		if err := v.runTun2socks(); err != nil {
 			log.Println(err)
 			return err
 		}
+		isTRunning = true
 
 		log.Printf("EnableLocalDNS: %v\nForwardIpv6: %v\nDomainName: %s",
 			v.EnableLocalDNS,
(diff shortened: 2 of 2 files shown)


AndroidLibV2rayLite/CoreI/Status.go
```diff
@@ -6,6 +6,7 @@ import (
 
 type Status struct {
 	IsRunning       bool
+	IsTRunning      bool
```
floatingrain, line 9: 命名不应该使用“意义不明确的缩写”，这点各大编码规范中都有提及。建议
```suggestion
	IsTun2socksRunning bool
```

AndroidLibV2rayLite/interact.go
```diff
@@ -111,6 +111,10 @@ func (v *V2RayPoint) GetIsRunning() bool {
 	return v.status.IsRunning
 }
 
+func (v *V2RayPoint) GetIsTRunning() bool {
```
floatingrain, line 114: 同样的，建议完整拼写命名

floatingrain (changes requested):
命名问题

AndroidLibV2rayLite/CoreI/Status.go
rurirei: do not use variable too long.

floatingrain: > do not use variable too long.

Maybe "IsT2SRunning" is better?

yuhan6665:
why do we need this bool? who will call GetIsTRunning()?

rurirei:
> more status check is needed, as #728

you could do next to handle possible errors on Tun2socks start of escort.go. @yuhan6665

yuhan6665:
> > more status check is needed, as #728
> 
> you could do next to handle possible errors on Tun2socks start of escort.go. @yuhan6665

#728 is a bit weird where looks like tun2socks error out after 10 seconds. Though I can't reproduce this issue. I think we need tun2socks log to know what happened in his device.
Even when there is an error in tun2socks, I think isTun2socksRunning will be true right?

rurirei:
> > > more status check is needed, as #728
> > 
> > 
> > you could do next to handle possible errors on Tun2socks start of escort.go. @yuhan6665
> 
> #728 is a bit weird where looks like tun2socks error out after 10 seconds. Though I can't reproduce this issue. I think we need tun2socks log to know what happened in his device.
> Even when there is an error in tun2socks, I think isTun2socksRunning will be true right?

yes you need to generate error from v.runTun2socks(), as it returns nil always.

Issue #865, opened 2021-01-29
V2rayNG new storage design


yuhan6665:
Over the past year of development, I had some discussion with @2dust. A few shortcomings of the current storage have been identified:

- [DPreference](https://github.com/DozenWang/DPreference) is an obsolete implementation.
- DPreference redirect multi-process access to main process. As a result, it can occupy a lot of redundant memory even when user is not looking at the MainActivity.
- Currently all server list data is stored under one key “ang_config”. This was ok for simple display the list in MainActivity. As the business logic grow, it become very inconvenient for updating single entry, like test results.
- Other useful feature, like sort, grouping, filtering, search is more difficult with current storage.

I plan to use [MMKV](https://github.com/Tencent/MMKV) to replace the DPreference. (AndroidX Datastore was also considered but I don’t believe it support multi-process access) I have done some profiling. MMKV shows good performance with similar memory usage, though it will increase APK size.

Currently v2rayNG has only one sharedPreference xml file. I plan to break it logically into 5 different files. The side benefit is that with true multi-process support from MMKV, it makes the file locking with smaller granularity.

`Key --> value`
1.	Main:
`ang_config --> [Id1, Id2..]` // Id is current guid. Duplicated server config is still allowed, I plan to add a new isSameServer() function to help with identify duplicated servers in the future
`pref_v2ray_config --> Id` //current selected server, plan to get rid of some of the redundant fields, like `pref_v2ray_config_domain`, `pref_v2ray_config_name`

2.	Server: (this is server data usually updated with subscription. Expect few changes)
`Id --> gson.toJson { remarks, configType, configVersion, subId, gson.toJson { V2rayConfig.OutboundBean}, addedTime }` //`addedTime `is a new field used to help sorting in the future
// in case of custom config, the gson.toJson will be full V2rayConfig. Reader will need to check configType before parse the json

3.	Server affiliation info: (this is server related data that potentially change very frequently, currently only testResult, consider to add traffic statistics, server health score in the future)
`Id --> gson.toJson { testResult }`

4.	Subscription:
`SubId --> gson.toJson { remarks, url, enabled, addedTime }`  //`addedTime `is a new field used to sort subscription

5.	Settings:
I plan to keep the default setting shared preference as is, but map the needed key value to MMIKV
pref_mode
pref_per_app_proxy
pref_per_app_proxy_set
pref_bypass_apps
pref_speed_enabled
pref_sniffing_enabled
pref_proxy_sharing_enabled
[...]

PR #965, opened 2021-03-27
New storage

yuhan6665:
https://github.com/2dust/v2rayNG/issues/865

@2dust Everything is done as far as I can tell. I tried to do all kinds of manual testing but since the impact is huge there might be missed bugs. I wish we have tester for this project :)
During testing, I notice some differences compare to before:
```
 + Save almost half memory when start from quick buttons
 + Test results are always saved
 + Custom config is now parsed and auto formatted
 - Custom config new fields in future version of core might need update in json schema, otherwise they are not saved
```


2dust:
https://github.com/2dust/v2rayNG/issues/865

PR #1146, opened 2021-07-16
fix update subscription error

jiuqianyuan:
Prevent server list from being emptied when update subscription failed or the subscription url is invalid #992


V2rayNG/app/src/main/kotlin/com/v2ray/ang/ui/MainActivity.kt
```diff
@@ -417,6 +417,7 @@ class MainActivity : BaseActivity(), NavigationView.OnNavigationItemSelectedList
                         URL(url).readText()
                     } catch (e: Exception) {
                         e.printStackTrace()
+                        return@launch
                         ""
```
yuhan6665: Thanks for your pr. Looks like after your return statement, the "" in line 421 is useless and should be cleaned up

jiuqianyuan: OK，Clean now

yuhan6665:
Looks good to me

PR #1169, opened 2021-08-03
indonesian language

fsteamID:
add this Indonesian to this project, Indonesian users are very enthusiastic about using this v2rayNG application


2dust:
Please use English language

PR #1204, opened 2021-08-23
Add support in subscription url for http basic authentication.

douo:
Add support in subscription url for http basic authentication. In order to obtain the resources correctly from the URL of the following form:

    http://username:password@example.com/


yuhan6665:
Seems good to me

PR #1420, opened 2022-03-21
更改captive portal链接

Yuzuki616:
使用google的captive portal的话会导致国内服务器测试失败，更换成cloudflare的captive portal可以避免这种情况


yuhan6665:
We might need to change it to a configurable setting. Since this will change user experience as well. Previously, when test return success, people know the proxy is working to circumvent GFW.

Yuzuki616:
> We might need to change it to a configurable setting. Since this will change user experience as well. Previously, when test return success, people know the proxy is working to circumvent GFW.

OK, that's better. This is also the best solution in my opinion. But i'm so sorry, i don't have the ability to implement  it.

PR #1589, opened 2022-07-10
Update README.md

mxbhaee:
Added the windows 11 process to build an .aar library for android


mxbhaee:
The ios build definitely needs mac os it won't build on windows

cuynu:
> The ios build definitely needs mac os it won't build on windows

V2rayNG only available on Android, its never available for iOS

PR #2010, opened 2023-02-14
build tun2socks  with ndk

blackier:
build tun2socks  with ndk


blackier:
verfied, and solve death loop while tun2socks launch failed.

rurirei:
not necessary i thought.

blackier:
> not necessary i thought.

This is a more convenient and understandable way to use tun2sock instead of just using a binary.

2dust:
Please PR the V2RayVpnService.kt part alone

Issue #2058, opened 2023-03-01
The config list should be sorted in reverse order


pouriaksrvi:
The current ascending order of config list is not recommended by UX design principles.
Also, it is inefficient.
Because, when the list is too long, the user should scroll to the end of the list to access newly added configs.
Or the user should run a test until the end of the list, and then click sort to bring the newly added configs (which are healthy) to the top of the list.
If new configs were added to the top of the list instead of its bottom, the user wouldn't need to wait for the tests to be completed. The user could click sort button once the tests reached the 20th or 30th config. i.e. before testing all list items. Testing half of the list (or even less) is enough when the new configs are at the top of the list and get tested first.
The new added config could be tested first, instead of last. So the user wouldn't need to complete the test.
In most of the cases, the user needs its latest added configs to be clicked instantly without needing to scroll to the end of the list.
Sorting by date added in descending order is better than ascending.
In long lists, adding a new config is a nightmare. Because the user can't wait for a complete test to sort. Scrolling is hard too. (imagine a list with 1000 configs or more)

### Expected Behavior
The latest added configs should be upper in the list and should be tested first. The list should be sorted in descending order.
So the user can click on sort, after testing first 20 items of the list and not waiting for the other items to be tested.
### Actual Behavior
The first added configs are upper in the list and get tested first. The list is sorted in ascending order.
When a new config gets added it goes to the bottom of the list.
The user needs to scroll to the end of the list.
Also, for sorting the 20 or 30 healthy items which are at the top of the list, and the proxies added a moment ago, the user should wait for the test controller to test the entire list while it is not needed to test all items if the added proxies were at top. (so the user could click on sort, before testing more than half of the list)

### Reproducibility method
1. Add too many proxies so that your list becomes too long so that waiting for test to be done to the end is timely.
2. Wait a few days until their quality decreases, their ping value increases, and some of them become unavailable.  
3. Add some new healthy and fresh proxies.
4. After running a test you should wait for a lot of time until the test process reaches them, so you can sort to make them go up.
5. Or you should scroll manually to the end (a too long list) to click and connect them.
6. 4 and 5 are both hard to do. The proxies are not at the top of the list so the user can't click on sort before finishing the test.

pouriaksrvi:
Look at how other proxy managers order their list. For example look at the internal proxy manager of Telegram Desktop as a popular one. The order is not ascending. Its the reverse.

pouriaksrvi:
[...]

PR #2072, opened 2023-03-06
[Issue Fixes]: #2058, #2076

pouriaksrvi:
1. [75af126](https://github.com/2dust/v2rayNG/pull/2072/commits/75af126102c0e9fd47de7697a6e64ce6c2598d45):
https://github.com/2dust/v2rayNG/issues/2058
**After this fix:**
Adding new configs (servers) cause them being on top of the list because most of the users, a moment after adding a new config, need to see it at the top, to click it.
**Before this fix:**
The user was forced to scroll a long list after adding each new config, or clear the list frequently before adding new configs to be able to access and click them.
Also, batch tests would never reach the latest added config when the list was too long.
The first added config (the oldest one, which is useless and expired or dead in most of the cases), was always staying at row 0 at the top of the list with no reason.

2. [01777db](https://github.com/2dust/v2rayNG/pull/2072/commits/01777db88430b431c863a434f9bd3a268b298996)
https://github.com/2dust/v2rayNG/issues/2076
**After this fix:**
More concurrent tests can be ran for tests. Test concurrency has no CPU usage practically because it is IO burst.
Having concurrent tests only lets the tests not being queued or locked or waited for each other.
The fixed 10 threads which in previous versions was being used, still exists. But more cached threads can get added to.
**Before this fix:**
Tests was being queued while there were no need to do this.

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayTestService.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayTestService.kt
@@ -15,7 +15,7 @@ import libv2ray.Libv2ray
 import java.util.concurrent.Executors
 
 class V2RayTestService : Service() {
-    private val realTestScope by lazy { CoroutineScope(Executors.newFixedThreadPool(10).asCoroutineDispatcher()) }
+    private val realTestScope by lazy { CoroutineScope(new ThreadPoolExecutor(10, Integer.MAX_VALUE, 60L, TimeUnit.SECONDS, new SynchronousQueue<Runnable>()).asCoroutineDispatcher()) }
 
     override fun onCreate() {
         super.onCreate()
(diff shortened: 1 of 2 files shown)


pouriaksrvi:
Some edits:
There were some typos in the pull request title and body texts.
I had used the word **"commit"** instead of **"config"**, many times.

daryo-sh4030:
برنامه روشن نمیشود

daryo-sh4030:
برنامه روشن نمیشود

daryo-sh4030:
برنامه روشن نمیشود

daryo-sh4030:
چرا روشن نمیشود

pouriaksrvi:
May I know the problem with this pull request?
The first commit (https://github.com/2dust/v2rayNG/issues/2058) is very necessary.
At least all users need it and lack of it is very annoying.

Issue #2116, opened 2023-03-21
Put last added configs at the top, instead of the bottom


pouriaksrvi:
With UX design principles, in most of the GUI applications, the newer items of tables and lists are shown at the beginning, to be accessible first.
For example:
1. In messengers chat page, you'll see latest messages first (At page bottom).
2. When you open chat list, you'll see the last updated chat, not the most inactive chat.
3. _Telegram Desktop_ **proxy list** sort
4. _Github_ Issue list page (the newer issues are shown first).

And a lot more examples...

In v2rayNG:
In most of the times, the user needs to click on the config which has been added a moment ago, not the oldest added config.
With the current time-ascending order, **the oldest added item stays at row 0 forever**, and added configs are put at the end of the list so the user needs to scroll to the end each time after adding new configs to click them.
When the list is too long, scrolling gets hard, so this problem becomes worse, and the user should remove old proxy configs frequently to shorten the list to overcome this problem. In other words, **users need do proxy management** instead of leaving this task for the app!!!
Also, making the user frequently shorten the list, increases the risk of **lockups** because there's more chance for all configs to get expired or died at once, when the config list is not long enough.

So reverse the default initial order please!

There's also **one more reason** for doing this: The oldest config, in most of the cases is expired or died. Why should it stay in front of the user's eye forever? New configs should be placed at the top and older configs should get shifted to down each time the user adds a new one. So they get hidden from user's eyes when the list is long.

pouriaksrvi:
This is very necessary.
Without this, the users should delete all earlier added configs each time they wanna add some new configs.
This is one of the most basic features which such an app should have.
It's strange that the incorrect order still exists, despite being too problematic.

pouriaksrvi:
I guess it is also a good idea to have a feature for pinning some configs to top

pouriaksrvi:
This issue has been fixed.

PR #2130, opened 2023-03-26
[New Feature] UX enhancement - Reversed the order of server list

pouriaksrvi:
Fixed incorrect default order of server list as described in the issue https://github.com/2dust/v2rayNG/issues/2116
The default order was problematic for users who add new servers more frequently

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/AngConfigManager.kt
@@ -756,6 +756,7 @@ object AngConfigManager {
 
             var count = 0
             servers.lines()
+                    .reversed()
                     .forEach {
                         val resId = importConfig(it, subid, removedSelectedServer)
                         if (resId == 0) {
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/MmkvManager.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/MmkvManager.kt
@@ -46,7 +46,7 @@ object MmkvManager {
         serverStorage?.encode(key, Gson().toJson(config))
         val serverList = decodeServerList()
         if (!serverList.contains(key)) {
-            serverList.add(key)
+            serverList.add(0, key)
             mainStorage?.encode(KEY_ANG_CONFIGS, Gson().toJson(serverList))
             if (mainStorage?.decodeString(KEY_SELECTED_SERVER).isNullOrBlank()) {
                 mainStorage?.encode(KEY_SELECTED_SERVER, key)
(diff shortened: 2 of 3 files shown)


cuynu:
tks i will add in my project

pouriaksrvi:
> tks i will add in my project

Sure, you're welcome. Add it ASAP please. It is necessary for all users but it is even more necessary for users who live in regions where having a long proxy list is needed.

pouriaksrvi:
Read the **reviews in Google play** store.

There are some people complaining "Why the app works very well but it doesn't work after a month"
The problem is they are not aware of the reason, which I have explained in the issue https://github.com/2dust/v2rayNG/issues/2116 therefore they don't know the new servers which they add, get added to the bottom of the list. They just see the old servers which their test results are -1 at the top of the list and the list is too long to see its bottom end which healthy servers are added by them.

2dust:
thanks for your pr
Consider the case of batch import and subscription import？

pouriaksrvi:
> thanks

You're welcome. 

> batch import and subscription import

Yes.
And the both of them have been tested successfully.

2dust:
I mean subscription should not be in reverse order

pouriaksrvi:
> subscription should not be in reverse order

Imagine these servers:
1, 2, 3, 4, 5, 6, 7, 8, 9
Having 3, 4, 5, 6 imported from a subscription, having the order you suggested it will be like:
1, 2, **_6, 5, 4, 3_**, 7, 8, 9

Isn't that inconsistent? Imagine the user scrolling the list from the beginning to the end;
In the middle of the list, subscription-imported servers get seen sorted in the opposite order.
I'm sure it makes the user confused.

A list is sorted by time in descending order if and only if all its sub-lists are sorted in the same order.

I can change it if it is needed. Are you sure?

pouriaksrvi:
The same reason for batch import:
Most of the times when the user copies servers from webpages or other sources, the newest servers get placed at the end of the list in the clipboard.
Therefore, they should be imported with the same order they have been copied, so the newest copied server gets put at the top of the list.
As explained in https://github.com/2dust/v2rayNG/issues/2116, the newest servers of the list, always have more chance to be healthy and working, so they should be the most accessible, in the default import order.

pouriaksrvi:
I have added a bug fix commit, fixing subscription import order.
Now it works in the way you suggested.

pouriaksrvi:
For batch import from clipboard, it is needed to import it in reversed order.
Imagine a webpage with this content:
```
post 1, 2020-02-01:
    some servers
post 2, 2020-02-02:
    some more servers
post 3, 2020-02-03:
    some more new servers
(more posts...)
```
Everyday, the older servers stop working or get blocked by the firewall and the admin provides some new servers in a new post.
The user randomly selects 100 recent posts by dragging the mouse cursor to copy some servers.
The latest posted servers should be imported to the top of the list because they are more likely to be alive, in tests and sorts.

Maybe users should not do it this way but they do.

wangkong1122:
谁会节点设置教程教教我谢谢大哥

Issue #2272, opened 2023-05-09
Per-release and Latest wait different packege name 


peymannorozimay:
Please provide the pre-release with a different package, because Google Play does not allow installation, and if auto update is enabled, it will reinstall the latest version.

Issue #2057, opened 2023-03-01
URI scheme for v2rayNG and v2rayN & share profile


APT-ZERO:
Hello
please add URI scheme for v2rayNG and v2rayN
v2rayng:// or v2rayn://

v2rayng://addprofile?name=...&url=...
v2rayng://addcfg?...

And option to share a profile with qrcode or copying it's v2rayng URI
Thank you

EDIT :
i found :
```
                <data android:scheme="v2rayng"
                    android:host="install-config" />
```
it's great, just need another host to add subscriptions using v2rayng scheme

PR #2273, opened 2023-05-10
[Feature] separate package names for pre release and dev builds

hadi-norouzi:
#2272 
For development you can run dev flavor that doesn't install on google play version and
when you build for pre release you can build pre_release flavor.

#2057 
import subscription from deeplink with ```v2rayng://install-sub/``` this pattern.

### DIFF
--- a/V2rayNG/app/build.gradle
+++ b/V2rayNG/app/build.gradle
@@ -36,6 +36,21 @@ android {
         }
     }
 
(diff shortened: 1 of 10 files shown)


donnyxray:
I currently import subscription URL's using v2rayng://install-config/ This works well. What makes v2rayng://install-sub/ different?

PR #2841, opened 2024-02-14
Update dependency

Malus-risus:


2dust:
Thanks for PR
But automatically updating dependencies can cause some problems

Malus-risus:
> Thanks for PR But automatically updating dependencies can cause some problems

Please check the action in my repository, I have debugged it, and the program has also been tested and can run successfully.

PR #2886, opened 2024-02-29
Addressing issue #2880

MrMalekfar:
Addressing the issue #2880 Importing Custom Configurations from Sub-Link Results in Incorrect Display of Integer Values as Floats #2880 I added  custom serialiser, because GSON by default parse number as Double.


MrMalekfar:
Sorry for old commits. this is my first PR request.

PR #2892, opened 2024-03-02
Added remarks to all custom configs

vfarid:
Added remarks to all custom configs in order to list by their original name instead of timestamp. This includes import/export of custom configs and those added by subscription links.

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/ui/ServerCustomConfigActivity.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/ui/ServerCustomConfigActivity.kt
@@ -91,7 +91,7 @@ class ServerCustomConfigActivity : BaseActivity() {
         }
 
         val config = MmkvManager.decodeServerConfig(editGuid) ?: ServerConfig.create(EConfigType.CUSTOM)
-        config.remarks = binding.etRemarks.text.toString().trim()
+        config.remarks = v2rayConfig.remarks ?: binding.etRemarks.text.toString().trim()
         config.fullConfig = v2rayConfig
 
         MmkvManager.encodeServerConfig(editGuid, config)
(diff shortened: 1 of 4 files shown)


TheOzs:
Grazie

mohammad20101368:
🙏

FransW5:
Is it possible to have multiple custom-configs and simple configs (vless/vmess) at the same time in the subscription link?
If not, can you correct this?
```
vless://
vless://
[{customConfig1}, {customConfig2}, {customConfig3}]
vmess://...
vmess://...
```

vfarid:
> Is it possible to have multiple custom-configs and simple configs (vless/vmess) at the same time in the subscription link? If not, can you correct this?
> 
> ```
> vless://
> vless://
> [{customConfig1}, {customConfig2}, {customConfig3}]
> vmess://...
> vmess://...
> ```

Its not a good idea imo, instead you can use two different subs, one for server configs and another one for custom configs.

V2rayNG/app/src/main/kotlin/com/v2ray/ang/ui/ServerCustomConfigActivity.kt
```diff
@@ -91,7 +91,7 @@ class ServerCustomConfigActivity : BaseActivity() {
         }
 
         val config = MmkvManager.decodeServerConfig(editGuid) ?: ServerConfig.create(EConfigType.CUSTOM)
-        config.remarks = binding.etRemarks.text.toString().trim()
+        config.remarks = v2rayConfig.remarks ?: binding.etRemarks.text.toString().trim()
```
2dust, line 94: Because this is manual editing, there is no need to do this.

vfarid: Its not manual edit, i set remarks from custom config's remarks field in its json if any, otherwise set timestamp for its name. This way configs having remarks in related json will show by their remarks in the list, instead of a random timestamp.

vfarid: If you ignore this part, editting custom config's remark inside the app will have no effect on the config name in the list. Btw its not that important.

sbh96:
how we can set remark name in subscription link?
is has any parameter to set config name?
@vfarid

vfarid:
> how we can set remark name in subscription link? is has any parameter to set config name? @vfarid

For each single config you can set remarks in json:
```json
{
  "inbounds": [...],
  "outbounds": [...],
  "remarks": "..."
}
```

PR #2915, opened 2024-03-09
Update dependency

Malus-risus:


savyjs:
Fixes #2929
Fixes #2932

Issue #2726, opened 2024-01-05
performance issue (not responding | freezing UI)


mosayeb-a:
- there is nothing any related issue.

when you go to the MainActivity, the ui freezes for almost a second. and if you go to the another screen and go back to the MainAcitvity, the app again freezes for a second. this freezing issue related to the reloadServerList function on the OnResume method on the MainActivity.

A solution to solve this problem is to load the servers asynchronously. But this solution has a problem that the servers are gradually added to the list, and to solve this we can show a loading view until all servers are added to the list.

If this solution is suitable for you, I can solve it.

https://github.com/2dust/v2rayNG/assets/108512702/b6522aeb-cc50-4b55-b42d-115a8f921046

device specification: 
- android 7
- Asus Zenfone Max Plus

**I am waiting for your response.**

PR #2916, opened 2024-03-09
fix freezing UI in MainActivity, #2726 

mosayeb-a:
when the reloadServer fun is called every time which onResume is called, the same servers reload again. so it cause performance issue.


2dust:
If the background service updates the server list, how to refresh the interface?

mosayeb-a:
I wrote a quick fix to solve this issue. But I think this project needs a lot of refactoring. Other solutions can be implemented which consume much less memory(RAM) or resourse  and are maintainable, but these solution require a lot of refactoring.

now it's look like this:

https://github.com/2dust/v2rayNG/assets/108512702/cceda59f-e688-48ce-b22d-2d29935337c1

mosayeb-a:
Why do you close it without merging it? It's a bad fix ? Please give more
information.

mosayeb-a:
Could you give me more explanation about background service which can change
the server list?

PR #2937, opened 2024-03-16
Free configs feature

mosayeb-a:
I fetch a list of free configs from this repository and show it in the UI. I am sure that many users will love this feature.

just like this:

https://github.com/2dust/v2rayNG/assets/108512702/3c89ce62-bfc6-485a-9e39-a3f87d9aafee

this is the **repo** : https://github.com/mrvcoder/V2rayCollector


2dust:
Thanks for PR
But the function is not in the plan

PR #2960, opened 2024-03-28
Decoupling the context from the view model 

Amir-yazdanmanesh:
- Move BroadcastReceiver from ViewModel to View
- Move Toasts handling from ViewModel to View layer

I've noticed that MainViewModel is tightly coupled with the Android Context, which goes against the best practice of keeping ViewModels independent of any Android framework components. I understand the importance of making ViewModels agnostic to the UI layer for better testability and reusability. I'm open to restructuring my codebase to remove any dependencies on the Context and separating out any view-related logic into appropriate view components.

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/ui/MainActivity.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/ui/MainActivity.kt
@@ -198,6 +198,18 @@ class MainActivity : BaseActivity(), NavigationView.OnNavigationItemSelectedList
         mainViewModel.isRunning.value = false
         startListenBroadcast()
         MessageUtil.sendMsg2Service(application, AppConfig.MSG_REGISTER_CLIENT, "")
+
+        mainViewModel.showMessageById.observe(this){ message ->
+            toast(message)
+        }
+        mainViewModel.deleteServerCount.observe(this){
+            toast(
+                getString(
+                    R.string.title_del_duplicate_config_count,
+                    it
+                )
+            )
+        }
     }
     fun startListenBroadcast() {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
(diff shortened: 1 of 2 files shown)


yuhan6665:
When I was an Android developer, I had many debates around this one. There is a school of thought that you should make business layer as pure java/kotlin and only do Android logic in View layer. I'd against that unless you are really reusing your code in other platforms. Otherwise you just create more overhead of development/runtime cost.
For example, you know your Activity can be killed and recreate any time, why put background logic in there?
Why does Android team introduced https://developer.android.com/reference/androidx/lifecycle/AndroidViewModel

Amir-yazdanmanesh:
Google Strongly recommended that :
ViewModels shouldn't hold a reference to any Lifecycle-related type. Don't pass Activity, Fragment, Context or Resources as a dependency. If something needs a Context in the ViewModel, you should strongly evaluate if that is in the right layer.

[https://developer.android.com/topic/architecture/recommendations , Vi](https://developer.android.com/topic/architecture/recommendations#viewmodel)

And also said : Do not use AndroidViewModel

why recommended  that ?
Google said:

A ViewModel usually shouldn't reference a view, Lifecycle, or any class that may hold a reference to the activity context. Because the ViewModel lifecycle is larger than the UI's, holding a lifecycle-related API in the ViewModel could cause memory leaks.

[developer.android.com/topic/libraries/architecture/viewmodel](https://developer.android.com/topic/libraries/architecture/viewmodel#implement-viewmodel)

Amir-yazdanmanesh:
Google Strongly recommended that :
ViewModels shouldn't hold a reference to any Lifecycle-related type. Don't pass Activity, Fragment, Context or Resources as a dependency. If something needs a Context in the ViewModel, you should strongly evaluate if that is in the right layer.

[https://developer.android.com/topic/architecture/recommendations , Vi](https://developer.android.com/topic/architecture/recommendations#viewmodel)

And also said : Do not use AndroidViewModel

why recommended  that ?
Google said:

A ViewModel usually shouldn't reference a view, Lifecycle, or any class that may hold a reference to the activity context. Because the ViewModel lifecycle is larger than the UI's, holding a lifecycle-related API in the ViewModel could cause memory leaks.

[developer.android.com/topic/libraries/architecture/viewmodel](https://developer.android.com/topic/libraries/architecture/viewmodel#implement-viewmodel)

PR #2966, opened 2024-04-01
Update dependency

Malus-risus:


V2rayNG/build.gradle.kts
```diff
@@ -1,6 +1,6 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 plugins {
-    id("com.android.application") version "8.2.2" apply false
-    id("com.android.library") version "8.2.2" apply false
+    id("com.android.application") version "8.3.1" apply false
```
2dust: The project is using an incompatible version (AGP 8.3.1) of the Android Gradle plugin. Latest supported version is AGP 8.2.2
See Android Studio & AGP compatibility options.

PR #2971, opened 2024-04-01
Upgrade dependency management

tim06:


2dust:
May I ask what is the point of doing this?

PR #3013, opened 2024-04-15
Implement update cache asynchronously

Amir-yazdanmanesh:
I added the 'updateCache' function in the Coroutine scope. If we want to load a lot of servers, the app will freeze.


MrIbrahem:
Actually, The app freeze every time I update servers

Issue #3100, opened 2024-05-09
Color bug


Fatifan:
Hello, the navigation buttons are not visible when the light theme is active.
![155915_v2rayNG](https://github.com/2dust/v2rayNG/assets/125193909/8f792250-ead3-4403-b5a3-a1f123c990aa)

PR #3108, opened 2024-05-11
optimize(ui):adjust color

FranzKafkaYu:
adjust color  for better UI,related Issue：https://github.com/2dust/v2rayNG/issues/3100

### DIFF
--- a/V2rayNG/app/src/main/res/values/colors.xml
+++ b/V2rayNG/app/src/main/res/values/colors.xml
@@ -8,6 +8,6 @@
     <color name="color_fab_inactive">#9C9C9C</color>
     <color name="color_secondary">#727272</color>
 
-    <color name="colorPrimary">#F5F5F5</color>
+    <color name="colorPrimary">#747474</color>
     <color name="colorAccent">#000000</color>
 </resources>


Fatifan:
I think #747474 is too dark and the black color (connection status) is not visible on it, I suggest #cdcdcd .

2dust:
Everyone has different aesthetics
Please provide a picture after you have modified it

Fatifan:
@2dust
 Do you believe that this is not a bug? If not, I'm pretty sure that #cdcdcd is a good choice for the "colorPrimary" tag and the black text and white icons will be visible on it but I'm not able to modify the code and provide a picture of the result, but you can try this.

FranzKafkaYu:
This PR is to solve the problem that the navigation bar at the bottom is less clearly visible.the following picture is the latest V2rayNG version in my XiaoMi Rednote 11 Pro+：
image

Fatifan:
@FranzKafkaYu 
I know and I had written #3100 myself, but the color you suggested in this PR is too dark and the black text above the icons cannot be visible on it.

FranzKafkaYu:
> @FranzKafkaYu I know and I had written #3100 myself, but the color you suggested in this PR is too dark and the black text above the icons cannot be visible on it.

I used AVD to test,the black text seems all right,just like this：

PR #3148, opened 2024-05-22
Find & set best fragment settings for the operator.

vfarid:
Added a new option in main menu to find and set best fragment options for selected config.
Bellow, you may watch the test result.
https://github.com/2dust/v2rayNG/assets/23278340/439e41f8-9d6f-400a-a9e2-e75d55a73ee8

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/AppConfig.kt
@@ -127,4 +127,7 @@ object AppConfig {
     const val MSG_MEASURE_CONFIG = 7
     const val MSG_MEASURE_CONFIG_SUCCESS = 71
     const val MSG_MEASURE_CONFIG_CANCEL = 72
+    const val MSG_MEASURE_FRAGMENT = 8
+    const val MSG_MEASURE_FRAGMENT_SUCCESS = 81
+    const val MSG_MEASURE_FRAGMENT_CANCEL = 82
 }
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/dto/FragmentsCache.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/dto/FragmentsCache.kt
@@ -1,0 +1,4 @@
+package com.v2ray.ang.dto
+import com.v2ray.ang.dto.V2rayConfig.OutboundBean.OutSettingsBean.FragmentBean
+data class FragmentsCache(val fragmentBean: FragmentBean,
+                          var ping: Long = -2)
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayTestService.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayTestService.kt
@@ -6,6 +6,9 @@ import android.os.IBinder
 import com.v2ray.ang.AppConfig.MSG_MEASURE_CONFIG
 import com.v2ray.ang.AppConfig.MSG_MEASURE_CONFIG_CANCEL
 import com.v2ray.ang.AppConfig.MSG_MEASURE_CONFIG_SUCCESS
+import com.v2ray.ang.AppConfig.MSG_MEASURE_FRAGMENT
+import com.v2ray.ang.AppConfig.MSG_MEASURE_FRAGMENT_CANCEL
+import com.v2ray.ang.AppConfig.MSG_MEASURE_FRAGMENT_SUCCESS
 import com.v2ray.ang.util.MessageUtil
 import com.v2ray.ang.util.SpeedtestUtil
 import com.v2ray.ang.util.Utils
@@ -32,7 +35,17 @@ class V2RayTestService : Service() {
                     MessageUtil.sendMsg2UI(this@V2RayTestService, MSG_MEASURE_CONFIG_SUCCESS, Pair(contentPair.first, result))
                 }
             }
+            MSG_MEASURE_FRAGMENT -> {
+                val contentPair = intent.getSerializableExtra("content") as Pair<Int, String>
+                realTestScope.launch {
+                    val result = SpeedtestUtil.realPing(contentPair.second)
+                    MessageUtil.sendMsg2UI(this@V2RayTestService, MSG_MEASURE_FRAGMENT_SUCCESS, Pair(contentPair.first, result))
+                }
+            }
             MSG_MEASURE_CONFIG_CANCEL -> {
+                realTestScope.coroutineContext[Job]?.cancelChildren()
+            }
+            MSG_MEASURE_FRAGMENT_CANCEL -> {
                 realTestScope.coroutineContext[Job]?.cancelChildren()
             }
         }
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/ui/MainActivity.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/ui/MainActivity.kt
@@ -355,6 +355,10 @@ class MainActivity : BaseActivity(), NavigationView.OnNavigationItemSelectedList
             mainViewModel.reloadServerList()
             true
         }
+        R.id.set_best_fragment_settings -> {
+            mainViewModel.findBestFragmentSettings()
+            true
+        }
         R.id.filter_config -> {
             mainViewModel.filterConfig(this)
             true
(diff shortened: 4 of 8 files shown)


MortezaJamii:
🙏

tempereddev:
Good 👌

pooyafo:
👍❤️

parsibox:
👍❤️

aliahmadiphoto4:
❤️‍🔥❤️‍🔥💌

jafarinejadsaeed:
👍❤️👍❤️👍

masoud-shahsavari:
useful 👍👍👍

sinagrimm69:
👍❤

bolbash:
I need it very much👌👍👍

Moratagfz:
❤️👍

fooadkhan:
❤️❤️❤️❤️❤️❤️❤️❤️👍👍👍👍👍

fooadkhan:
👍👍👍

fooadkhan:
❤️❤️

fooadkhan:
👍👍

ghanad:
👍

mrvcoder:
Nice 👀

xDragonZ:
Why you guys want to spam all these unrelated comments?

boredcircuit:
The community has shown mixed results and sentiments regarding the impact of fragmentation. If it necessitates frequent changes, it could likely serve as a marker for traffic patterns or IP flagging and potentially aiding attackers. I believe it's good to be cautious about the `Widespread` availability of such traffic pattern adjustments, as it may have unintended consequences.

essi90:
Good

vfarid:
> The community has shown mixed results and sentiments regarding the impact of fragmentation. If it necessitates frequent changes, it could likely serve as a marker for traffic patterns or IP flagging and potentially aiding attackers. I believe it's good to be cautious about the `Widespread` availability of such traffic pattern adjustments, as it may have unintended consequences.

The main and only use of tlshello fragment is to hide SNI. It is not related to ip flagging and it also cant help attackers! The only reason we need to find best options is because there is different behavior in different operator during the time. Everything is about caching and its costs. Proper caching in firewall will affect fragment and may disable it at the end.

2dust:
This function will not be added to the main program of v2rayNG
You can fork a copy and add this function or develop a separate App to implement this function.

Similar functions such as this  https://github.com/2dust/v2rayNG/discussions/2970

vfarid:
Hey @2dust,
I really dont understand why you ignore new features to the app while they are usefull and have no negative effect on the whole app.
I know finding the clean IP is usefull in china as well as iran since my other seperate scanner is used in china. It is the same for fragment and i hope you revise your decision.
By the way if you are not happy to add new working features to this app, i stop sending futher PR and will respect your descision.

Thank you.

yuhan6665:
@2dust I had doubt when RPRX and you merged fragmentation feature. Now that I see it is widely used in Iran. Maybe this pr is ok as long as the code is clean as a separate module.

javidshah2:
> @2dust I had doubt when RPRX and you merged fragmentation feature. Now that I see it is widely used in Iran. Maybe this pr is ok as long as the code is clean as a separate module.

Please add that, the dictatorial rule of the Islamic Republic causes the death of many freedom-loving people of Iran every day, either by using guns or the high cost and lack of medicine and thousands of other indirect methods. For us, having a free internet is like fighting against the murderers of the Iranian people, we destroy them and bring our own voice to the world, although if no one wants to help, it is our duty to be the voice of our own people. For us, free internet is like needing bread and water, just like you can't live without minimum food, without free internet, it means easy genocide of Iranians without anyone knowing. Each of the new features acts like an oxygen device that does not allow more people to be massacred.

🙏🏻🙏🏻🙏🏻

mablue:
Some very useful updates just ignored because:
### "We Don't Know why"
I also don't know anything about internet quality in china but, Ignoring people time(with ignoring new updates to connect to the free internet) is actually like ignoring people lives in Iran.
But there is a tricky thing people in Iran can't realize:
# No water can be drawn from an empty well, even if we send the bucket deeper, faster or more times to the well.

These apps are designed to protect user data by setting a different protocol on top of normal protocols with maybe 1 or maximum 5 servers in each profile...

But we are using 1000 - 5000 V2Ray servers in each phone to test and go out of this shit for some 
seconds or if we are lucky for some minutes.
And it's not the true way

Thank @vfarid thanks @2dust I hope you undrestand what happening to our people in iran and Feel more sympathy with us. It is inevitable that there are many comments and requests. If you are very upset, assume that we are like monkeys(example: last comment) in a cage that are shouting, «take us with you» & «dont live us alone» & etc ...

## bad news
We currently do not feel the need for the security of our data in Iran.(You can find many modded apps in most famous iranian open websites like farsroid and p30download ) and you can check Apk files with tools like VirusTotal to see all of them have many warms trojans and viruses...but we use them cuz play-store and many other services are banned in iran from inside and outside. So we use these apps like V2RayNG Hiddify-Next and NekoBox etc... just for having free access to internet. we know privacy is important but not in a traditional prison.

Mahdi-zarei:
> > The community has shown mixed results and sentiments regarding the impact of fragmentation. If it necessitates frequent changes, it could likely serve as a marker for traffic patterns or IP flagging and potentially aiding attackers. I believe it's good to be cautious about the `Widespread` availability of such traffic pattern adjustments, as it may have unintended consequences.
> 
> The main and only use of tlshello fragment is to hide SNI. It is not related to ip flagging and it also cant help attackers! The only reason we need to find best options is because there is different behavior in different operator during the time. Everything is about caching and its costs. Proper caching in firewall will affect fragment and may disable it at the end.

The idea to hide the SNI is based on the false belief that the GFW needs the SNI to prevent a connection to a server which is behind a CDN, but what the GFW needs is a "flag" that distinguishes a connection from the mainstream, and fragmentation is a very obvious flag, as fragmentation of client hello is a very rare event and thus the GFW can simply close any connection that has its client hello fragmented. Just like how SSH tunnel works because they have not decided to block it, configurations which use fragmentation can easily be blocked by simply applying a rule which drops any connection which has the beginning part of a client hello but it's incomplete, and this would have a very small and negligible disruption on the mainflow of the internet.

APT-ZERO:
I can't SSH to my servers because of SSH Tunnel which made ISPs to ban SSH on most of IP Ranges
And now I can't update my subscription that is behind cloudflare cdn, because of this fragment sh!t
I think it's better to stop this Fragment sh!t to prevent cloudflare ip ranges getting blocked completely

At least limit the global fragment option of v2rayNG to Cloudflare IP Ranges only to prevent our direct and relay IPs getting detected by firewall (as no normal app tries to connect to any IP with fragmented tlshello or first packets)

yuhan6665:
@APT-ZERO you know you can add a simple filter on your server, just checking first packet length is probably ok

mohammad20101368:
❤️

boredcircuit:
> The main and only use of tlshello fragment is to hide SNI. It is not related to ip flagging and it also cant help attackers! The only reason we need to find best options is because there is different behavior in different operator during the time. Everything is about caching and its costs. Proper caching in firewall will affect fragment and may disable it at the end.

I believe there is a misunderstanding regarding the concept of fragmentation. Caching is not required, and a flag in the IP header will be present whether desired or not; otherwise, the packet won't be reassembled at the destination. If there are mixed results with different configurations as though they are blocking or attempting to block, your modifications started to prevent fragmentation from occurring(no fragment in physical connection), whether due to shorter timeframes or longer packets and perhaps that's the secret sauce to its helpfulness.

> The idea to hide the SNI is based on the false belief that the GFW needs the SNI to prevent a connection to a server which is behind a CDN, but what the GFW needs is a "flag" that distinguishes a connection from the mainstream, and fragmentation is a very obvious flag, as fragmentation of client hello is a very rare event and thus the GFW can simply close any connection that has its client hello fragmented. Just like how SSH tunnel works because they have not decided to block it, configurations which use fragmentation can easily be blocked by simply applying a rule which drops any connection which has the beginning part of a client hello but it's incomplete, and this would have a very small and negligible disruption on the mainflow of the internet.

It's not for us to determine whether it's beneficial to have certain options, but if an obvious flag has been left and even proved useful for a time, it may have aided in analysis and the creation of whitelist rules (which they can do it again and some will keep providing them more info with that flag time to time). Therefore, I'm indifferent to whose hands it falls into or through what distribution channel it passes. However, when such logic is incorporated into an open-source project and attackers are aware, it's foolish to underestimate them. A long-term perspective is better, especially since the situation is more severe than in other countries, with regular users contributing to spam and attackers potentially risking more false positives.

Fatifan:
@vfarid I suggest that you add an option in the settings to select the country and when selected, more settings for that country will be added.

PR #3384, opened 2024-07-30
Add localhost DNS support

mayampi01:


2dust:
There is no problem with the code, can you tell me if you have tested it and it works?

mayampi01:
> There is no problem with the code, can you tell me if you have tested it and it works?

I tested it with custom config, no problem.

dyhkwong:
Why did you think this will work? `localhost` is a `net.Resolver` wrapper. It means using system resolver, with cgo this is the DNS addresses assigned to VpnService, so this causes resolving loop. Did you really test it with VpnService?

mayampi01:
> Why did you think this will work? `localhost` is a `net.Resolver` wrapper. It means using system resolver, with cgo this is the DNS addresses assigned to VpnService, so this causes resolving loop. Did you really test it with VpnService?

1.8.36 is out, if you don't have Android Private DNS enabled, you can test it after disabling local DNS.

When using Android Private DNS, the `localhost` refers to the Android Private DNS.

Without Android Private DNS and local DNS, when the VPN DNS is set to IPv4/v6 address, the `localhost` refers to the VPN DNS.

Without Android Private DNS and local DNS, when the VPN DNS is set to "localhost" (or other non-IP string), the `localhost` refers to the DNS servers of the default network.

`localhost` resolving loop only occurs when using local DNS without Android Private DNS.

dyhkwong:
DNS servers of VpnService can no way be a non-IP address https://developer.android.com/reference/android/net/VpnService.Builder#addDnsServer(java.net.InetAddress)
So it just gets ignored by the app if you set "VPN DNS" to a non-IP address. You mean ignorance of non-IP string is the expected behaviour of your design?

> localhost resolving loop only occurs when using local DNS without Android Private DNS.

yay that's one of the cases. In most cases localhost is not the DNS of the underlying network.

mayampi01:
I made a mistake, when enabling local DNS or VPN DNS, to use Android Private DNS with `localhost`, the `hosts` is also need to specify.
https://github.com/2dust/v2rayNG/pull/3433

mayampi01:
> DNS servers of VpnService can no way be a non-IP address https://developer.android.com/reference/android/net/VpnService.Builder#addDnsServer(java.net.InetAddress) So it just gets ignored by the app if you set "VPN DNS" to a non-IP address. You mean ignorance of non-IP string is the expected behaviour of your design?
> localhost resolving loop only occurs when using local DNS without Android Private DNS.
> yay that's one of the cases. In most cases localhost is not the DNS of the underlying network.

https://github.com/2dust/v2rayNG/blob/2574553180ea640f67e03962b0921ed08d91c4b1/V2rayNG/app/src/main/kotlin/com/v2ray/ang/util/Utils.kt#L145

If `localhost` is not suitable for you, just don't use it.

This is for those who need to refer to Android Private DNS or VPN DNS or the DNS servers of the default network.

dyhkwong:
> This is for those who need to refer to Android Private DNS or VPN DNS or the DNS servers of the default network.

Okay this makes sense.

PR #3388, opened 2024-07-31
routing: First block, then direct, and finally proxy

mayampi01:
Make the `block` really work, otherwise most of the `block` rules will be `proxy` or `direct` first.

And first `block`, then `direct`, and finally `proxy` is also the official recommended order.

https://xtls.github.io/document/level-0/ch08-xray-clients.html


2dust:
https://github.com/2dust/v2rayNG/commit/bbdee92f3733637075f3d7372a217c34e4c4602d

PR #3407, opened 2024-08-02
Refactor removeDuplicateServer Function to Simplify Duplicate Removal

codewithtamim:
Refactored the `removeDuplicateServer` function in `MainViewModel` to simplify the logic for identifying and removing duplicate servers. The updated implementation now efficiently filters out duplicates and removes them from both the server list and cache, improving code readability and maintainability.


2dust:
`removeDuplicateServer `modified for worse performance.
Because the second loop is more times

codewithtamim:
> `removeDuplicateServer `modified for worse performance.
> Because the second loop is more times

😭

PR #3413, opened 2024-08-02
Updated getLocale function to use appropriate Locale constants

codewithtamim:
Updated getLocale function to use appropriate Locale constants


codewithtamim:
All prs submitted from yesterday's  work. Thanks

PR #3422, opened 2024-08-05
Use localhost DNS server to make geosite:private work

mayampi01:
`geosite:private` contains many domain names that need to be resolved to the local router (gateway) and can only be resolved using the `localhost` DNS server.

Such as:
asusrouter.com
routerlogin.com
hiwifi.com
miwifi.com

https://github.com/v2fly/domain-list-community/blob/master/data/private

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/AppConfig.kt
@@ -107,7 +107,7 @@ object AppConfig {
 
     /** DNS server addresses. */
     const val DNS_PROXY = "1.1.1.1"
-    const val DNS_DIRECT = "223.5.5.5"
+    const val DNS_DIRECT = "localhost"
     const val DNS_VPN = "1.1.1.1"
 
     /** Ports and addresses for various services. */


2dust:
`const val DNS_DIRECT = "223.5.5.5"`
It's just a default value, you can set it to `localhost `yourself.

If it is changed to `localhost `, it will affect many Chinese users

mayampi01:
How will changing the default DNS server from `223.5.5.5` to `localhost` (for example, 192.168.1.1) affect many Chinese users?

2dust:
Because `localhost `will be polluted by the carrier's dns on Chinese users' devices.

mayampi01:
`223.5.5.5` as a UDP DNS query can also be polluted by ISPs.
China Mobile and China Broadnet will hijack and respond to UDP DNS query.
If you want to avoid pollution, you should use HTTPS (`https+local://223.5.5.5/dns-query`) or QUIC (`quic+local://223.5.5.5`, may be supported in the future).
TCP (`tcp+local://223.5.5.5`) DNS pollution is less common, but it is still possible.

[基于旁路抢答机制的异网DNS管控实践.pdf](https://github.com/user-attachments/files/16490964/DNS.pdf)
[基于镜像抢答机制的异网DNS重定向系统在广电宽带中的应用.pdf](https://github.com/user-attachments/files/16490971/DNS.pdf)

PR #3433, opened 2024-08-07
Hardcode popular Android Private DNS rule to fix localhost DNS problem

mayampi01:
This is a temporary solution. The final solution should be to add user-defined hosts function.


mayampi01:
https://github.com/2dust/v2rayNG/pull/3384#issuecomment-2273136544

PR #3450, opened 2024-08-10
Fix ABI filter retrieval in applicationVariants.all block

codewithtamim:
Updated the `applicationVariants.all` block in the build.gradle file to use the new API for filtering outputs. Replaced deprecated `getFilter("ABI")` method with `filters.find { it.name == "ABI" }?.value` and used `filterIsInstance` for type safety.

### DIFF
--- a/V2rayNG/app/build.gradle.kts
+++ b/V2rayNG/app/build.gradle.kts
@@ -57,30 +57,27 @@ android {
         }
     }
 
-    applicationVariants.all {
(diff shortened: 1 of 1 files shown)


ghost:
Screenshot_2024-08-10-16-34-22-66

@CodeWithTamim 
Error❗

codewithtamim:
Yeah I know. Will fix it

PR #3452, opened 2024-08-10
Fix: Suppress UnspecifiedRegisterReceiverFlag warning

codewithtamim:
Suppress UnspecifiedRegisterReceiverFlag warning

Added @Suppress("UnspecifiedRegisterReceiverFlag") annotation to resolve the warning related to the unspecified register receiver flag.


2dust:
There is no need to add this pr, it is better to have a warning that can prompt developers to

Issue #3475, opened 2024-08-13
name app


elifian:
Hello,

I would like to request a change to the button label in the application dropdown. Currently, the button is labeled "Переключить" (russian language), but I would prefer it to be updated to "v2rayNG".

This change would help in better identifying the application

Thanks for the work done
<img src="https://github.com/user-attachments/assets/3bd72d87-c8d2-4944-8fd4-5c4deacdc047" width="300" />

2dust:
<string name="app_widget_name">Переключить</string>
    <string name="app_tile_name">Переключить</string>

Please pr fix

PR #3485, opened 2024-08-15
Fix issue #3475

codewithtamim:
Fix issue where the label in the tile and widget was not displaying as `v2rayNG` for Russian language users


solokot:
Perhaps the text in the English translation should also be changed, since it is now similar to the changed text?
https://github.com/2dust/v2rayNG/blob/c0fda6fcba8ba026338bee634642c8c63adb2a43/V2rayNG/app/src/main/res/values/strings.xml#L4
https://github.com/2dust/v2rayNG/blob/c0fda6fcba8ba026338bee634642c8c63adb2a43/V2rayNG/app/src/main/res/values/strings.xml#L5

PR #3517, opened 2024-08-25
Correct the capitalization of protocol names

xixu-me:


codewithtamim (commented):
I re checked it, works good. but but don't change the GitHub yml

xixu-me:
I was careless, I restored the GitHub yml

codewithtamim:
> I was careless, I restored the GitHub yml

good job

PR #3593, opened 2024-09-24
Fix invalid VPN DNS

HeXis-YS:
尝试修复#3586。
无论是否启用本地DNS都使用设置中提供的DNS。
如果DNS请求以UDP发送，可以由dnsgw转发至Xray-core的本地DNS服务器，其他协议的DNS请求以常规流量的方式进入Xray-core。

### DIFF
--- a/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt
+++ b/V2rayNG/app/src/main/kotlin/com/v2ray/ang/service/V2RayVpnService.kt
@@ -137,16 +137,12 @@ class V2RayVpnService : VpnService(), ServiceControl {
             }
         }
 
-        if (settingsStorage?.decodeBool(AppConfig.PREF_LOCAL_DNS_ENABLED) == true) {
-            builder.addDnsServer(PRIVATE_VLAN4_ROUTER)
-        } else {
-            Utils.getVpnDnsServers()
-                .forEach {
-                    if (Utils.isPureIpAddress(it)) {
-                        builder.addDnsServer(it)
-                    }
+        Utils.getVpnDnsServers()
+            .forEach {
+                if (Utils.isPureIpAddress(it)) {
+                    builder.addDnsServer(it)
                 }
-        }
+            }
 
         builder.setSession(V2RayServiceManager.currentConfig?.remarks.orEmpty())
 


t-e-s-tweb:
Isn't this totally unsafe? Since requests will be routed through ipv4/v6 without the use of DNS over https or tls? Currently all requests are going through that and local DNS isn't even being used on ws or splithttp with this commit.

HeXis-YS:
@t-e-s-tweb 
I'm not sure what you're trying to say. Let me explain what I understand, and please correct me if I'm wrong.

First of all, this PR aims to solve a DNS leak issue.

Initially, when the local DNS server is enabled, v2rayNG uses PRIVATE_VLAN4_ROUTER, which is 26.26.26.2, as the DNS server. In earlier versions, DNS requests were sent over the traditional UDP protocol. These traditional DNS requests are forwarded to the local DNS server port (usually 10853) defined by the user in Xray-core via the dnsgw module of badvpn-tun2socks. In Xray-core, requests received on the local DNS server port are processed according to a special mechanism. The original destination DNS server is ignored and the query is sent to the DNS server specified by the user in the Xray configuration instead.

However, the way these DNS requests are sent may be controlled by the system or certain dependent libraries. In newer versions/systems, these DNS requests may be sent over TLS/HTTPS/QUIC for “security” reasons. badvpn-tun2socks does not recognize these requests, and they are forwarded to the socks port of Xray-core as regular traffic. If the user has not set the corresponding routing rules in Xray-core, these requests will really be sent to 26.26.26.2. This is not only insecure, but you are unlikely to receive a response.

Therefore, in this case, the VPN DNS servers should be set to some real available DNS servers, which is what this PR does. It doesn't do anything to break compatibility. If you notice anything not working, then most likely they never worked in the first place.

Let's get back to your security concerns.

First of all, here we can only specify which DNS server to use, but not decide on the specific protocol to use. If your system decides to use UDP, there is nothing we can do to change it, at least I don't know how.

Second, theoretically, all DNS requests will enter the Xray-core, except for applications that are not proxied. Therefore, whether these DNS requests will be leaked depends entirely on the user's configuration. Here you can configure the DNS server to be used, the protocol (DoT/DoH/DoQ) and the routing of traffic as you like.

Finally, the existence of a local DNS server is not necessary. I think it is very likely that the v2rayNG code was inherited from a client from shadowsocks-era. It inherited the code and its historical legacy. Shadowsocks may use the dnsgw module of badvpn-tun2socks to redirect DNS requests, and v2rayNG simply reuses this function. In my opinion, this is unnecessary. You can completely separate DNS traffic from all traffic received on the socks port by configuring Xray-core, without having to use a separate port.

t-e-s-tweb:
@HeXis-YS sorry if I couldn't explain it better before. 

Let me give you an example of my use case,  previously when I set a ad-blocking DNS server, all DNS requests were being forwarded to that so only that DNS server showed on DNS leak test websites such as ipleak.net or browserleaks.com/ip. 

Ads were being blocked and the DNS that I set in local DNS Remote option in v2rayng was being used. 

Now, no matter which DNS I specify in remote DNS , that isn't being used. Websites only see the VPN DNS that is set in VPN settings in v2rayng app. 

So the options for remote DNS in v2rayng app isn't taking effect. 

Kindly set this ( https://dns.adguard-dns.com/dns-query ) as your remote DNS and help me to make it work.

HeXis-YS:
@t-e-s-tweb Your settings are taking effect. If you explicitly specify a DoH server in your v2rayNG settings (not in your Xray-core configuration), they will be forwarded as regular traffic to the socks port.

PR #3649, opened 2024-10-03
Update strings.xml

mh292929:
Translation of untranslated strings in Persian language


codewithtamim:
thank you I forgot to do that when adding :)

PR #3716, opened 2024-10-20
Organize locale

codewithtamim:
Organize locale


2dust:
Is this better to create an enum type?

codewithtamim:
Maybe not :)

codewithtamim:
Okay lets do with enum

codewithtamim:
Used enum now check

PR #3730, opened 2024-10-21
Update build.yml

ghost:


codewithtamim:
@2dust bruh android api is now 35 with this pr. it should be 21 check before merging

PR #3752, opened 2024-10-23
Update libs.versions.toml

ghost:


ghost:
mmkvStatic version 2.0.0 has removed the 32 bit arm variant

PR #3785, opened 2024-10-25
add repo size badge

codewithtamim:
add repo size badge


2dust:
That's not very useful.

codewithtamim:
👍

PR #3798, opened 2024-10-28
Optimize display effect

ghost:
Add postDelayed optimize vertical black stripes on right side when open items in the sidebar


2dust:
感谢PR

PR #3827, opened 2024-11-02
Update AppManagerUtil.kt

aksSuzumiya6666:
Update AppManagerUtil.kt
* Fixed an error when compiling set compile Sdk=35


2dust:
applicationInfo If it is equal to null, what to do?

aksSuzumiya6666:
I'm sorry, I didn't consider this issue.

PR #3919, opened 2024-11-11
Delete quic protocol

ghost:


2dust:
The next version will be processed, and it cannot be processed in a minor version.

PR #3942, opened 2024-11-14
Update V2rayConfig.kt

ghost:


codewithtamim:
Does it work ? Have you tested with normal and custom config ?

ghost:
> Does it work ? Have you tested with normal and custom config ?

I did this on a guess, if possible, try it, tell me if there is a problem, and I will fix it

codewithtamim:
> > Does it work ? Have you tested with normal and custom config ?
> 
> I did this on a guess, if possible, try it, tell me if there is a problem, and I will fix it

Bro, just to confirm, this program is moving to production, right? So, as contributors, we should try to keep it as bug-free as possible. If you’re making any changes, could you please test them first to ensure everything works smoothly? You can find split HTTP configs from some websites to test with, so that might be helpful. Thanks so much for your efforts!

ghost:
> > > Does it work ? Have you tested with normal and custom config ?
> > 
> > 
> > I did this on a guess, if possible, try it, tell me if there is a problem, and I will fix it
> 
> Bro, just to confirm, this program is moving to production, right? So, as contributors, we should try to keep it as bug-free as possible. If you’re making any changes, could you please test them first to ensure everything works smoothly? You can find split HTTP configs from some websites to test with, so that might be helpful. Thanks so much for your efforts!

My goal is to update splithttp, the boss said that when it becomes stable, they will add it just like it was added to v2ran today. Thank you for your review.
I will close the pr 
I am a beginner and I don't have much access. Thank you for understanding me

PR #3943, opened 2024-11-14
Update gradle-wrapper.properties

ghost:


codewithtamim:
You don't need to update Gradle daily unless prompted by Android Studio. Most developers typically use a common Gradle version, and frequent updates can force them to download new versions constantly, which may slow down the development process.

ghost:
> You don't need to update Gradle daily unless prompted by Android Studio. Most developers typically use a common Gradle version, and frequent updates can force them to download new versions constantly, which may slow down the development process.

That's right, because this version was stable, I pr it

PR #3972, opened 2024-11-18
Added shared object binary for hysteria2

codewithtamim:
Added shared object binary for hysteria2


codewithtamim:
source https://github.com/lry127/Android-Hysteria-2-client

2dust:
use this hysteria-plugin
https://github.com/dyhkwong/Exclave/releases

PR #3976, opened 2024-11-18
Update persian translate

ghost:
Update persian translate


2dust:
Pull the latest and PR

PR #3991, opened 2024-11-19
Import rulesets from qrcode

Chocolate4U:
Add capability to import rulesets from QRcode

Note: I have reused `resetRoutingRulesetsFromClipboard` function for importing rulesets from QRcode. Therefore I've renamed this function to `resetRoutingRulesets` to be more inline with it's usage. Also old `resetRoutingRulesets` function renamed to `resetRoutingRulesetsFromPresets`.

Screenshot_20241119_234404-2

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/handler/SettingsManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/handler/SettingsManager.kt
@@ -43,12 +43,12 @@ object SettingsManager {
     }
 
 
-    fun resetRoutingRulesets(context: Context, index: Int) {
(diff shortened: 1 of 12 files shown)


2dust:
Thanks for your PR
But I am worried that qrcode cannot store large rules

Chocolate4U:
Yes, but for most use cases it's enough, for example the current "China Whitelist" preset has 1299 characters. (which is a mix of alpha-numeric, symbols and Chinese characters). I have successfully encoded up to 2232 characters of the same character-set. You can also encode more characters if you only use alpha-numeric characters and symbols (not using complex characters like Chinese or Persian)

PR #4030, opened 2024-11-24
Update dependencies 

ghost:
Add dependabot for workflow and gradle dependencies
Update dependencies gradle
This PR is appropriate because dependabot informed gradle dependencies and workflows dependencies
@2dust


ghost:
It makes your job easier to update dependencies. It is a very good thing if you decide to merge
@2dust

2dust:
You don’t need to always use the latest version, because each update needs to be tested and verified.

ghost:
> You don’t need to always use the latest version, because each update needs to be tested and verified.

See how dependabot works, it offers the new version. Do you want to like it or not? Anyway, it was a good thing, you can do whatever you like.

PR #4063, opened 2024-11-28
Update Luri Bakhtiari translation

hosseinabaspanah:


APT-ZERO:
I guess only 50 v2rayNG users uses this tribe language

hosseinabaspanah:
> I guess only 50 v2rayNG users uses this tribe language

Laughing out loud

First, do some research on the Luri Bakhtiari language, then comment on this...
Although the statistics on the number of speakers of this language have not been updated for a long time, I recommend you read this article...
https://en.m.wikipedia.org/wiki/Bakhtiari_dialect

PR #4077, opened 2024-11-30
Update translation 

ghost:


2dust:
`V2rayNG/app/src/main/assets/custom_routing_global ` This is for users in mainland China, no English required

PR #4108, opened 2024-12-08
Use system locale for string formatting

codewithtamim:
Use system locale for string formatting

https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/format.html
``` kotlin 
inline fun String.Companion.format(locale: Locale?, format: String, vararg args: Any?): String
```


2dust:
The places you modified are strings with specific functions。
should not be modified like this

Issue #4107, opened 2024-12-08
Hysteria2 pinSHA256


ReturnFI:
Hello, @2dust

The config generated by my script uses `pinSHA256`, which has a URI example like this:

`hy2://user:password@IP:PORT?obfs=salamander&obfs-password=password&pinSHA256=sha256/u.....4o=&insecure=1&sni=bts.com`

The issue is resolved when we remove `pinSHA256` from the config.
However, this parameter is necessary for verify the server's certificate fingerprint.

Thank you.

PR #4109, opened 2024-12-08
Fix Hysteria2 pinSHA256 validation and configuration (#4107)

codewithtamim:
Fix Hysteria2 pinSHA256 validation and configuration (#4107)


2dust:
The attribute values ​​of the configuration file should be ensured by the user, and the app does not make any special processing

PR #4111, opened 2024-12-08
Add all application initialization into initializeAng method

codewithtamim:
Add all application initialization into initializeAng method


2dust:
Initialization functions should not be placed in `Utils`

codewithtamim:
> Initialization functions should not be placed in `Utils`

Then where?

PR #4152, opened 2024-12-19
use different version for abis

UjuiUjuMandan:
Hi, have a look at this: https://developer.android.com/build/configure-apk-splits#configure-APK-versions

Google recommends different version codes for different ABIs, and the current way seems to be incompatible with F-Droid.

I ~~set it as starting from 4~~ added a plus 5000000  to avoid inconvenience for users.

### DIFF
--- a/V2rayNG/app/build.gradle.kts
+++ b/V2rayNG/app/build.gradle.kts
@@ -60,7 +60,7 @@ android {
     applicationVariants.all {
         val variant = this
         val versionCodes =
-            mapOf("armeabi-v7a" to 8, "arm64-v8a" to 7, "x86" to 6, "x86_64" to 5, "universal" to 4)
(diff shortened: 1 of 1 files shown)


UjuiUjuMandan:
Discussion in progress.

2dust:
Because Google Play need app bundle for publishing, all apks must be the same version  to be able to upgrade normally

UjuiUjuMandan:
Then Google must handle different version codes for ABIs; I will download a bundle to see.

2dust:
When publishing to Google Play, `use release.aab`, which already contains all architectures and has only one  versionCode;
Then when publishing to GitHub, there are 4 apks. For the convenience of upgrading, all versions use the same  versionCode

There is no problem using one  versionCodefor all abi, because only one apk will be installed on a mobile phone.

UjuiUjuMandan:
Thanks for your Information.

PR #4161, opened 2024-12-20
switch to quickie-foss for QR code

UjuiUjuMandan:
Step 2 in #4160.


codewithtamim (commented):
```
Unresolved reference: quickie
``` 
@UjuiUjuMandan .
you forgot to add that in versions catalog

PR #4164, opened 2024-12-21
Fix flavor build

UjuiUjuMandan:
Only the version code differs, the two flavors’ dependencies are all the same, generate either is ok.


UjuiUjuMandan:
Emmm. Wait, there are folders for 2 flavors.

PR #4166, opened 2024-12-21
Update build.yml

ghost:
github not use Flavors.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,9 @@ jobs:
       run: |
         cd ${{ github.workspace }}/V2rayNG
         chmod 755 gradlew
-        ./gradlew licenseFdroidReleaseReport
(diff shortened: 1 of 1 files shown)


UjuiUjuMandan:
@2dust Sorry, please revert this. It's necessary for Reproducible Builds.

The workflow is like this:

1. GitHub Actions builds and uploads the APKs of both flavors to releases, one is with `-fdroid` suffix, the other is not.
2. fdroidserver builds the APKs of `fdroid` flavors.
3. fdroidserver fetches your APKs with `-fdroid` suffix from GitHub Releases.
4. fdroidserver compares them. If passed, releases it with copying your signatures.

UjuiUjuMandan:
The `sed` patching is really unnecessary; I have only seen downstream patches of upstream; how opposite?

UjuiUjuMandan:
I'll revert that after the veery slow `diffoscope` completed.

ghost:
😂a little trouble, all right.

PR #4168, opened 2024-12-21
Change fdroid flavor applicationId

UjuiUjuMandan:
To avoid mix installing.

@2dust linsui suggested only preserving `fdroid` flavor in GitHub Releases to avoid confusing users. I'm not sure. What do you think?

### DIFF
--- a/V2rayNG/app/build.gradle.kts
+++ b/V2rayNG/app/build.gradle.kts
@@ -52,7 +52,7 @@ android {
     productFlavors {
         create("fdroid") {
             dimension = "distribution"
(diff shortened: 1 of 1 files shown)


2dust:
> To avoid mix installing.
> 
> @2dust linsui suggested only preserving `fdroid` flavor in GitHub Releases to avoid confusing users. I'm not sure. What do you think?

no
The number of fdroid users is a small part

UjuiUjuMandan:
Yes. I hope there won't be any genius asking why there are 2 v2rayNG on my phone!

PR #4167, opened 2024-12-21
Build and cache libtun2socks and libv2ray

UjuiUjuMandan:
- https://github.com/2dust/AndroidLibV2rayLite/pull/42

The latest build failed, probably try increasing heap size in `org.gradle.jvmargs`.

PR #4169, opened 2024-12-21
Ignore libtun2socks.so

UjuiUjuMandan:
Actuality it should be `*.so`. I'll build `libhysteria` in days.


codewithtamim:
It's needed as the workflow only builds the core

UjuiUjuMandan:
> It's needed as the workflow only builds the core

#4167.

PR #4171, opened 2024-12-22
Optimization logcat after change quickie-foss

ghost:
After changed quickie-foss, scanning cause lot of log. throttle log com.google.zxing.NotFoundException.


2dust:
It is better not to display` zxing.NotFoundException` information at all

UjuiUjuMandan:
This is an upstream problem.

UjuiUjuMandan:
They don't even enabled issue.

- https://github.com/T8RIN/QuickieExtended

~~I'll try fixing that.~~

PR #4172, opened 2024-12-22
Refine cache key

UjuiUjuMandan:
The former strategy only considered the changes of submodules, not including the changes of the build script, as in https://github.com/2dust/AndroidLibV2rayLite/pull/43. Changing it to checking commit header.


UjuiUjuMandan:
Emmm wait a moment.

UjuiUjuMandan:
OK, let it be as is. I cannot find any more elegant way.

PR #4175, opened 2024-12-22
Remove go patch

UjuiUjuMandan:
Continuing  https://github.com/2dust/v2rayNG/pull/3841#discussion_r1894860380.

Cache deleted and still no problem. I don't think it's necessary any more.


2dust:
There will be no compilation error, but an exception will occur during runtime

UjuiUjuMandan:
Fine.

PR #4181, opened 2024-12-23
Clarify flavors

UjuiUjuMandan:
Related:

- #4162
- #4168

Make sure that there's only one v2rayNG ~~and probably save some GA free minutes~~.

The `split` means it should be built to APKs, and the `playstore` is for AAB.

split versioning:

```kotlin
output.versionCodeOverride = (100 * variant.versionCode + versionCodes[abi]!!).plus(5000000)
```

playstore versioning:

```kotlin
output.versionCodeOverride = (100 * variant.versionCode + 5).plus(5000000)
```

versionCodes[abi] are all smaller than 5. As long as the Play Store channel is slower than GitHub Release, there will be no problem for users.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -131,26 +131,26 @@ jobs:
       run: |
         cd ${{ github.workspace }}/V2rayNG
         chmod 755 gradlew
-        ./gradlew licenseFdroidReleaseReport
-        ./gradlew assembleRelease -Pandroid.injected.signing.store.file=${{ steps.android_keystore.outputs.filePath }} -Pandroid.injected.signing.store.password=${{ secrets.APP_KEYSTORE_PASSWORD }} -Pandroid.injected.signing.key.alias=${{ secrets.APP_KEYSTORE_ALIAS }} -Pandroid.injected.signing.key.password=${{ secrets.APP_KEY_PASSWORD }}
+        ./gradlew licenseSplitReleaseReport
+        ./gradlew assembleSplitRelease -Pandroid.injected.signing.store.file=${{ steps.android_keystore.outputs.filePath }} -Pandroid.injected.signing.store.password=${{ secrets.APP_KEYSTORE_PASSWORD }} -Pandroid.injected.signing.key.alias=${{ secrets.APP_KEYSTORE_ALIAS }} -Pandroid.injected.signing.key.password=${{ secrets.APP_KEY_PASSWORD }}
 
     - name: Upload arm64-v8a APK
       uses: actions/upload-artifact@v4
       if: ${{  success() }}
       with:
         name: arm64-v8a
-        path: ${{ github.workspace }}/V2rayNG/app/build/outputs/apk/*/release/*arm64-v8a*.apk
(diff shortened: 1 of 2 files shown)


2dust:
We must ensure that the `versioncode `of `GitHub `and` Google Play Store` are consistent, so that users can easily replace them.
Because `Google Play Store` often changes its policies, we must ensure that there is a channel for updates.

2dust:
> probably save some GA free minutes.

We can reduce the apk, for example, only keep the arm64 and universal versions.
Modern mobile phones are basically arm64 architecture, other download universal version

UjuiUjuMandan:
> so that users can easily replace them.

I fully understand that and considered that.

I said no trouble after considered all possible scenarios .

1. Users installed the app via Play Store, and tries to install **same versionName** via GitHub later.
2. Users installed the app via Play Store, and tries to **update or install a higher versionName** via GitHub later.
3. Users downloaded and installed the APK from GitHub, but later try to install **same versionName** via Play Store.
4. Users download and install the APK from GitHub, but try to **update, aka install an higher versionName** via Play Store.

Take an example, the current versionCode in build.gradle is 622.

And in this patch, for arm64-v8a, it will be `100 * 622 + 1 +5000000`. Finally 5062201, and for playstore it’s 5062205.

What will happen if you released an update? It may increase to like 623 in build.gradle. So…

You released the newer APK (5062301) on GitHub first, and submitted newer AAB (5062305) for Google to review.

Now the newer versionName is on GitHub but not yet on Play Store.

- For 1, it would not happen, they always grabs the newest.
- For 2, it will succeed, 5062301 > 5062205.
- For 3, it will succeed, 5062205 > 5062201.
- For 4, it would not happen, there’s no newer version on Play Store yet!

And hopefully Google finished their review. Now the newer versionName is on both channels!

- For 1, it will definitely fail, 5062301 < 5062305, but no loss.
- For 2, it would not happen, there’s no newer version on GitHub yet!
- For 3, it will succeed, 5062305 > 5062301.
- For 4, it will succeed, 5062305 > 5062201.

And, if you didn’t know about it, Android checks for versionCode instead of versionName.

UjuiUjuMandan:
> You released the newer APK (5062301) on GitHub first

And probably not by You yourself, it can be uploaded to Releases by GA as soon as there were built.

UjuiUjuMandan:
> We can reduce the apk, for example, only keep the arm64 and universal versions.

It won't save any time if you keep universal, native code will still be compiled for armeabi-v7a and x86s, just did not packaged as a standalone APK.

The packaging and uploading take little time, but the build is not.

---

I just considered it twice, I was wrong actually. This patch won't save any time than before, since the only difference of two flavors is versionCode, so only packaging is involved.

2dust:
Your submission only changes the `versioncode `of the `PlayStore `, and the rest is basically the same as before.
Please keep the current `versioncode ` for PlayStore and GitHub. There is another reason that is not easy to mention at the moment, and we have to wait for the `PlayStore `policy.

UjuiUjuMandan:
The policy, emmm.. OK. Just take you time.

PR #4202, opened 2024-12-29
optimization build when commit upversion

ghost:


2dust:
`Setup release tag`
This step is not necessary, and manual publishing does not require much work.

ghost:
> `Setup release tag` This step is not necessary, and manual publishing does not require much work.

when commit is up 1.0.0 use it, it's automatic publishing. if just manual not need.If you don't need it, you can close it. Thank you.

PR #4208, opened 2024-12-29
disable bypass lan in vpn code.

patterniha:
xray-core can bypass private IPs and no need to bypass private IPs in vpn-code.

and just because geoip/geosite:private rule exist (and route to direct) is not a valid reason to bypass lan in vpn-code and **all other rules must be considered** .(if these rules exist xray-core bypass lan IPs by itself and no need to bypass lan in vpn code)

in addition **routing rules is for xray-core only** and should not used for vpn-code.
problems such as:
cannot block specific private ip.
cannot block specific port of private ips.
and ....
will occur if bypass all lan ips in vpn code.

and in the last if the bypass lan in vpn code is necessary(why?) you can put simple checkbox for it.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/handler/SettingsManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/handler/SettingsManager.kt
@@ -116,24 +116,9 @@ object SettingsManager {
         MmkvManager.encodeRoutingRulesets(rulesetList)
     }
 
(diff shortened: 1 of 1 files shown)


2dust:
A VPN setting will be added in the next version.

patterniha:
> A VPN setting will be added in the next version.

thank you very much.
mixing xray-core settings with vpn settings caused me a lot of problems.
i am waiting for new version.

PR #4226, opened 2025-01-03
Build and cache libhysteria2.so

UjuiUjuMandan:
- https://github.com/2dust/v2rayNG/pull/4169#issue-2754184596

Workflow passed in https://github.com/UjuiUjuMandan/v2rayNG/actions/runs/12595200811, and tested with my own hysteria2 proxy server, using Android Studio and the same libhysteria2.so built APK.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,8 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
+      with:
+        submodules: 'true'
 
     - name: Prepare build dir
       run: |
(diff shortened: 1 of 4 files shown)


UjuiUjuMandan:
Also, you might want to display the version of Hysteria build in About page.

UjuiUjuMandan:
Is this related at all?

1. The current libv2ray is built from xray-core, not the other one.
2. The build process is JimmyHuang454-free.

UjuiUjuMandan:
@2dust Hey, is the patch 600296 in AndroidLibXrayLite [workflow](https://github.com/2dust/AndroidLibXrayLite/blob/main/.github/workflows/main.yml##L22) also needed here? There's no problem running hysteria client on my device, but I'm not sure if other device, architecture or Android version would meet that error.

- https://github.com/2dust/v2rayNG/pull/4175#issuecomment-2558418835

PR #4249, opened 2025-01-09
Reproducible Builds for libhysteria2.so

UjuiUjuMandan:
- Build artifacts match - tested.
- Functionality - tested.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -97,6 +97,12 @@ jobs:
       with:
         go-mod-file: 'AndroidLibXrayLite/go.mod'
 
+    - name: Patch Go use 600296
+      #https://go-review.googlesource.com/c/go/+/600296
+      run: |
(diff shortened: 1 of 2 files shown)


UjuiUjuMandan:
Ah... Wait.

UjuiUjuMandan:
I'll open another pull request of switching build.yml to matrix to workaround a seemingly Android bug, see  [fdroid/fdroiddata#2998](https://gitlab.com/fdroid/fdroiddata/-/issues/2998#note_1405234894). But only after this one is merged, so that I don't need to solve all that conflict.

UjuiUjuMandan:
Emmm.... probably matrix is not needed, just disabling isUniversalApk is fine. I'll try if it will circumvent this bug.

UjuiUjuMandan:
No, it chose the `armeabi-v7a` versionCode for all ABIs instead.

PR #4253, opened 2025-01-11
matrix

UjuiUjuMandan:
See:

- https://github.com/2dust/v2rayNG/pull/4249#issuecomment-2582857540

The bug is like this:

When the filter contains than one ABI, all APK's some .dex files will get a same versionCode, their AndroidManifest.xml get the correct and according one though.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
       if: steps.cache-libtun2socks-restore.outputs.cache-hit != 'true'
       run: |
         cd ${{ github.workspace }}/AndroidLibXrayLite
-        bash compile-tun2socks.sh {{ matrix.abi }}
+        bash compile-tun2socks.sh ${{ matrix.abi }}
         tar -xvzf libtun2socks.so.tgz
       env:
         NDK_HOME: ${{ steps.setup-ndk.outputs.ndk-path }}
@@ -110,7 +110,7 @@ jobs:
     - name: Build libhysteria2
       if: steps.cache-libhysteria2-restore.outputs.cache-hit != 'true'
       run: |
-        bash libhysteria2.sh {{ matrix.abi }}
+        bash libhysteria2.sh ${{ matrix.abi }}
       env:
         ANDROID_NDK_HOME: ${{ steps.setup-ndk.outputs.ndk-path }}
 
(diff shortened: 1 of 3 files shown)


2dust:
I think you made the script too complicated.
Please don't put the ABI in this script

2dust:
libhysteria2's script is a standalone one.

UjuiUjuMandan:
> Please don't put the ABI in this script

Sorry?

UjuiUjuMandan:
> libhysteria2's script is a standalone one.

I'll specific the ABI in its command parameter instead.

Better to do this for libtun2socks too.

UjuiUjuMandan:
The universal APK will not be built or uploaded at all now.

UjuiUjuMandan:
> The universal APK will not be built or uploaded at all now.

It can still be built locally, if no gradlew parameter was provided.

2dust:
The `compile-tun2socks.sh` and `libhysteria2.sh` scripts should be run independently to complete their work. 
The scripts before this PR are good enough.
`build.gradle.kts` also needs to work independently without `github`, and the `ABI `in it should not be removed.

UjuiUjuMandan:
> The compile-tun2socks.sh and libhysteria2.sh scripts should be run independently to complete their work.

They will compile all ABIs, just like before if no parameter was provided.

> The scripts before this PR are good enough.

I don't think so, considering that bug is preventing it from being reproducible. And, building different ABIs in parallel will probably save some time.

> build.gradle.kts also needs to work independently without github, and the ABI in it should not be removed.

OK.

UjuiUjuMandan:
> The scripts before this PR are good enough.

OK, you're right at build.gradle.kts. I just reviewed that and got to know I don't need to change it at all.

UjuiUjuMandan:
I'll report this to Gradle instead.

UjuiUjuMandan:
Maybe you have misunderstood? It's unnecessary to change build.gradle, but the matrix is necessary... 

I suggest you to fix it here...the bug reporting system of AGP seems horrible.

2dust:
There is no misunderstood.
It's just that the release script is too complicated.
Or we can just not release universal.apk anymore.

UjuiUjuMandan:
> Or we can just not release universal.apk anymore.

Well, the god dammit… that bug is not only limited to universal, it eventually chooses a same one for all ABIs. For example, it uses the versionCode of armeabi-v7a for itself and all other 3 ones.

PR #4352, opened 2025-02-24
fixup libtun2socks cache

UjuiUjuMandan:
1. RPRX deleted badvpn `main` branch and replaced it with `new`, resulting in a missing cache key. Before I changed it to `new`, I found that the previous `/refs/heads/BRANCH_NAME` hack was unnecessary, I can just use HEAD as long as it's not submodule-in-submodule.
2. I forgot to change the cache path, it cannot find cache.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,8 +25,8 @@ jobs:
       id: cache-libtun2socks-restore
       uses: actions/cache/restore@v4
       with:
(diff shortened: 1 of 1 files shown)


UjuiUjuMandan:
https://github.com/UjuiUjuMandan/v2rayNG/actions/runs/13493530542/job/37695749290

PR #4513, opened 2025-04-18
real delay test

megasoheilsh:
Please make the delay test real. https has no correlation with ping and will mislead the user.


2dust:
你可以在设置中自己改

PR #4691, opened 2025-06-26
Showing Local Ip Address In Settings

UltraSina:
i added a text to settings that show user phone local ip address that make using proxy easier.


2dust:
感谢你的 PR
但是目前没有计划增加此功能

PR #4824, opened 2025-08-10
fix speed difference 

megasoheilsh:
https://github.com/2dust/v2rayNG/discussions/4809
https://github.com/megasoheilsh/v2rayNG/actions/runs/16857696097
The speed difference problem between saeeddev94, nebobox and v2rayng in hev mode was found, if the mtu value is increased, the speed problem will be solved.
If you think applying MTU will be problematic for all users, put it in a submenu.
thanks for supporting

photo_2025-08-10_09-06-03


2dust:
感谢，
MTU 应该需要增加一个设置，由用户自己选择或输入

PR #4869, opened 2025-08-24
Reasonable app sorting order

kvinty:
Order by selected status, then system/3rd party status, then by name


t-e-s-tweb:
Much better readability in long lists now! Thanks. 👍🏻

PR #4926, opened 2025-09-22
Fix XHTTP option processing

supernovaw:
~~XHTTP expects a `streamSettings.xhttpSettings.downloadSettings` object (with fields `address`, `port`, `realitySettings`, etc) that is mistakenly named `extra` instead of `downloadSettings`, which causes these settings to be ignored and have no effect.~~

~~This makes it impossible to achieve uplink and downlink separation.
You might want to additionally rename other parts of code accordingly, this is the least amount of change that fixed the problem for me.~~

~~P.S. this is my first time contributing to a FOSS project, yay! I hope you accept my fix :3~~


2dust:
https://github.com/XTLS/Xray-core/discussions/4113

supernovaw:
I'm so sorry! I somehow missed that `downloadSettings` goes in `extra`. Thank you for your quick response. I hope I didn't bother you.

PR #4991, opened 2025-10-31
Trigger workflow only on fastlane directory

UjuiUjuMandan:
Save some CI minutes...


UjuiUjuMandan:
Ah, sorry? We don't need to validate fastlane metadata at every push since no metadata was changed.

PR #5207, opened 2026-01-26
Fix: Stop service when app is removed from recents (Locked state)

sharkxuanbee:
Fixes an issue where the notification persists after swiping away the app, specifically when the app is 'Locked' in the Android task switcher.

Implements onTaskRemoved in V2RayVpnService and V2RayProxyOnlyService to ensure services effectively stop.


2dust:
在有些设备上，移除任务后台还在运行，这个是一个功能

PR #5332, opened 2026-02-28
Shadowsocks - support for query string, mux, and UoT

AlekseyMartynov:
For Shadowsocks protocol **with any stream settings configured** (non TCP transport, TLS, etc):
- Allow query string in `ShadowsocksFmt`
- Allow mux in `updateOutboundWithGlobalSettings`
- UDP over TCP flag is set to `true` if `method` is one of `2022-blake3-*` - comment in code explains why

These changes expose Xray-specific capabilities while preserving compatibility with standard / vanilla Shadowsocks configurations.


2dust:
感谢你的 PR
对 SS，我们不再继续增加功能了，请使用 VLESS

PR #5447, opened 2026-04-05
feat: build native libs through gradle script

### LINKED ISSUE #5446: [FEATURE] : Automate build process for hev tun and libXray using Gradle scripts.
Right now we run shell scripts by hand and copy outputs into app/libs, writing that into Gradle would let a normal build produce those native pieces automatically ! without separate manual steps needed.

codewithtamim:
resolves #5446, explained the problem statement there.


2dust:
感谢 PR
目前不准备接收，你可以在自己的分支使用。

PR #5458, opened 2026-04-07
Require SOCKS auth and rotate local credentials per service start

MrTeeett:
## Changes
- Enforce `auth: "password"` for local SOCKS inbound.
- Add SOCKS `accounts` support to inbound config DTO.
- Generate random local SOCKS username/password in memory per service start.
- Apply the same credentials to `hev-socks5-tunnel` config (`username`/`password`).
- Rotate credentials when the service starts/restarts.

## Security impact
Previously, local SOCKS could be used without auth by other apps on the device.
Now unauthenticated access is rejected.

PR #5459, opened 2026-04-07
Sock5 auth

Alekcy:
### This PR adds SOCKS5 authentication support for the local proxy and exposes it in Settings.

What changed:

- Added local SOCKS5 username/password support for the generated runtime config
- Applied the same auth handling to both Xray local inbound and hev-socks5-tunnel
- Added an auto mode that generates random local SOCKS5 credentials on service start
- Enabled auto-generated credentials by default
- Added a manual mode for custom credentials
- When switching from auto to manual mode, the current generated credentials are prefilled into the editable fields
image
image
image

### UPD!!!
Closing SOCKS5 isn't a complete security solution.
Even with split tunneling, any application can use tun0 and discover the real IP address.
You can verify this by running the following command in Termux:
`curl -s https://ifconfig.me/ip --interface tun0`


hippo2025:
Does it work with custom JSON configs?

Alekcy:
> Does it work with custom JSON configs?

It should now. I've updated it for custom JSON configs too, but it still needs testing.

hippo2025:
Does it take socks5 parameters from custom JSON or replaces it with its own generated password?

hippo2025:
There is also a parallel PR

hippo2025:
How about also picking port randomly?

Alekcy:
> Does it take socks5 parameters from custom JSON or replaces it with its own generated password?

replace. Authorization is added to all socks inbounds

Alekcy:
> How about also picking port randomly?

Added. By default, the port is randomly selected, but you can disable it and set your own.
image
image

hippo2025:
And for custom configs you also replace the specified port, right?

Alekcy:
Oh, I forgot about that. I'll add it later.

Alekcy:
Updated. Now the selected/generated port is applied to the custom config.

EVMIchigo:
How to use it?

soda-pop-ice-cream:
Scan 65k 127.0.0.1 ports on beefy phone is pretty easy task. Address setting should be added too, we have at least 127.0.0.0/8 and 169.254.0.0/16 subnets for this. FFS don't limit users with some /24

aeiouaeiouaeiouaeiouaeiouaeiou:
@Alekcy could you please release the binary files along with their VirusTotal scan results in your fork?

Alekcy:
I made a build for testing. But as it turns out, closing SOCKS5 isn't a solution.
Even with split tunneling, any application can use tun0 and discover the real IP address.
You can check this by running it in Termux
`curl -s https://ifconfig.me/ip --interface tun0 `

[v2rayNG-test_2.0.18_universal.apk](https://github.com/Alekcy/v2rayNG/releases/download/test/v2rayNG-test_2.0.18_universal.apk)
https://www.virustotal.com/gui/file/5e36a2e41e0c71773ad7a12119248baff96875931a557d9e423a3cccf3ebcc47/details

Alekcy:
And it looks like the problem with Tun0 can't be fixed at the V2RayNG level. So, as far as I understand, we need to look at xray-core or hev-socks5-tunnel reps

2dust:
服务启动时生成随机的本地 SOCKS5，这个是没有意义的。

重复 PR https://github.com/2dust/v2rayNG/pull/5458

w35th:
I have a kind of a solution for `curl -s https://ifconfig.me/ip --interface tun0` problem: you can create the second user in android OS (Settings -> System -> Users). If tun0 (VPN) was created in user1's space, apps from user2 won't be able to connect directly to tun0 (but they will be able to see its existence = vpn existense).

PR #5493, opened 2026-04-16
fix: send x-hwid for subscription updates

unins0:
## Summary
- send `x-hwid` with subscription update requests
- keep the value stable per install by persisting it once
- avoid per-request random IDs that can trigger false device-limit hits
## Why
Some subscription endpoints reject requests when device binding is enabled and `x-hwid` is missing.
Using a stable stub restores compatibility without adding hardware fingerprinting or changing behavior for servers that ignore the header.
## Notes
- applied only to subscription download requests
- no UI changes
- no real hardware identifiers are collected

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
@@ -46,6 +46,9 @@ object AppConfig {
     const val SUBSCRIPTION_AUTO_UPDATE = "pref_auto_update_subscription"
     const val SUBSCRIPTION_AUTO_UPDATE_INTERVAL = "pref_auto_update_interval"
     const val SUBSCRIPTION_DEFAULT_UPDATE_INTERVAL = "1440" // Default is 24 hours
+    // Remnawave HWID header (stable per install).
+    const val HWID_HEADER_NAME = "x-hwid"
+    const val PREF_FAKE_HWID = "pref_fake_hwid"
     const val SUBSCRIPTION_UPDATE_TASK_NAME = "subscription_updater"
     const val PREF_SPEED_ENABLED = "pref_speed_enabled"
     const val PREF_CONFIRM_REMOVE = "pref_confirm_remove"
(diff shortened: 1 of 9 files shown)


2dust:
请问这是哪些订阅源要求的？

unins0:
remnawave at least, maybe someone else has

hippo2025:
Maybe we should accept this but let users input custom HWID (in particular, the same HWID on multiple devices), as mentioned [here](https://github.com/2dust/v2rayNG/issues/5516). This way you can buy subscription but circumvent its restrictions. (Yes, I really don't like HWID and all restrictions of this sort)

unins0:
> Maybe we should accept this but let users input custom HWID (in particular, the same HWID on multiple devices), as mentioned [here](https://github.com/2dust/v2rayNG/issues/5516). This way you can buy subscription but circumvent its restrictions. (Yes, I really don't like HWID and all restrictions of this sort)

Added an option to send the x-hwid separately for each subscription, and a field showing the current x-hwid, which can be edited

unins0:
@2dust Why?

Ponywka:
@unins0 
> Why?

There's no need to defend the idiots at Remnaware and other "UI panels" who implemented this *pratecteon*. This protection is absolute bullshit.

unins0:
> @unins0
> 
> > Why?
> 
> There's no need to defend the idiots at Remnaware and other "UI panels" who implemented this _pratecteon_. This protection is absolute bullshit.

I’ve added a field where you can type in your HWID or paste it. I’m against this too, which is exactly why I decided to write this

ImMALWARE:
> This protection is absolute bullshit.

Yes, you are absolutely right, but why not just add support for importing that kind of subscriptions?

ImMALWARE:
@Ponywka, nobody here defends them, we just want support for these subscriptions in v2rayNG

Ponywka:
@ImMALWARE 

> @Ponywka, nobody here defends them, we just want support for these subscriptions in v2rayNG

The more mainstream client applications stop supporting this unnecessary feature, the more useless it becomes in the long run. I think it’s a form of protest against stupidity!

ImMALWARE:
@Ponywka, lol, are you serious? no, this form of protest (that in reality is the inconvenience of creating a server-layer that will add this header) is stupidity

if 2 apps will remove supporting this header, nobody will stop requiring it from the servers. users will just switch to happ and other non-opensource apps. is this your form of "protest"?

Ponywka:
@ImMALWARE 
> users will just switch to happ and other non-opensource apps

They already use it lol. Use fork clients or don't support such bad providers who uses this bullshit restrictions

PR #5509, opened 2026-04-19
Add negative UID handling to process routing

Kapkap5454:
Adds special item in process picker routing to handle process with negative UID (-1). Necessary to protect against malicious apps abusing SO_BINDTODEVICE which results in forced access to tun0 (linux kernel >=5.7).

This PR requires modification of libv2ray_android.go in another repo, to allow handling of negative UID.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
@@ -219,6 +219,8 @@ object AppConfig {
     const val REALITY = "reality"
     const val HEADER_TYPE_HTTP = "http"
 
+    const val UNIDENTIFIED_PACKAGE = "__unknown_app__"
+
     val DNS_ALIDNS_ADDRESSES = arrayListOf("223.5.5.5", "223.6.6.6", "2400:3200::1", "2400:3200:baba::1")
     val DNS_CLOUDFLARE_ONE_ADDRESSES = arrayListOf("1.1.1.1", "1.0.0.1", "2606:4700:4700::1111", "2606:4700:4700::1001")
     val DNS_CLOUDFLARE_DNS_COM_ADDRESSES = arrayListOf("104.16.132.229", "104.16.133.229", "2606:4700::6810:84e5", "2606:4700::6810:85e5")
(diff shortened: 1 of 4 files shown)


2dust:
此 PR 可能有点用处，可以让用户自己设置未知流量的路由行为，比如 direct 
但是风险是 UID = -1 的情况比较复杂，没有实测 xray 对于返回 -1 的行为如何处理？
请问是否实测过？

hippo2025:
> 此 PR 可能有点用处，可以让用户自己设置未知流量的路由行为，比如 direct 但是风险是 UID = -1 的情况比较复杂，没有实测 xray 对于返回 -1 的行为如何处理？ 请问是否实测过？

I think this indeed requires investigation. For instance, will it always be so, or maybe in a new Android version Google will change the behaviour? At the moment, I suggest merging this PR, so that it's easy for us to test this feature in real life and see how stable it is. What I'm worried about is that maybe sometimes Xray core will return -1 in some other case, maybe some temporary errors. In addition, I'm not fully convinced there are no other ways for spy apps to circumvent our restrictions... But only real life testing and further investigation will show this. At the moment, we can just leave this feature as experimental.

Also refer to this [commit](https://github.com/Exclude0122/xivpn/commit/c3093817bf0145dae332f7377d92cb89fad4123c), I assume @Exclude0122 has done some testing already.

Kapkap5454:
> 此 PR 可能有点用处，可以让用户自己设置未知流量的路由行为，比如 direct 但是风险是 UID = -1 的情况比较复杂，没有实测 xray 对于返回 -1 的行为如何处理？ 请问是否实测过？

Hi, 2dust!

Yes, I tested this build yesterday, and it works correctly.

From what I gathered, xray doesn't care if its -1 or other, it just works with whatever is passed in the config in routing. I tested passing -1 and I tested 0 (root). Both were handled by xray just fine.

The only thing that stops this from working is this piece of code in your other repo [AndroidLibXrayLite](https://github.com/2dust/AndroidLibXrayLite/tree/main) /libv2ray_android.go

if uid < 0 {
    return 0, "", "", fmt.Errorf("processFinder, not found for %s %s:%d -> %s:%d", network, srcIP, srcPort, destIP, destPort)
}

I deleted it and built new .aar file. And it worked. Maybe you want to handle it differently other than simply deleting.

This is the first time I write anything for android. But code changes are minimal, so I gave it a shot.

Let me do some tests once again, just for safety. And then I think it can be merged.

2dust:
https://github.com/2dust/v2rayNG/actions/runs/24628375342

你们试一试效果怎样吗?

Kapkap5454:
> https://github.com/2dust/v2rayNG/actions/runs/24628375342
> 
> 你们试一试效果怎样吗?

It definitely works. I see no errors or anything, its just working. Thank you, 2dust!

PR #5526, opened 2026-04-21
Add profile download speed test

Hohlas:
## Summary

This PR adds a lightweight download speed test for profiles.

v2rayNG already has useful latency checks, but latency does not always reflect real profile quality. Some servers respond quickly while having poor throughput. This feature adds a practical second signal: measured download speed through the selected profile.

The new `Speed test` action first runs the existing real delay test, then tests download speed only for profiles that passed. Results are shown next to the delay value in the profile list.

## Screenshot

Speed test results in profile list

## Highlights

- Adds a new `Speed test` menu action
- Runs real delay before download speed testing
- Tests only profiles with successful delay results
- Displays speed next to delay in the profile list
- Adds configurable speed test URL, timeout, and concurrency
- Uses temporary core instances and local proxy download
- Keeps the implementation isolated and easy to adjust

## Why

Real delay is useful, but it is not always enough to choose the best server. A profile can have low latency and still provide poor download performance. This feature helps users compare profiles by real usable throughput.

## Testing

Tested on a real Android device using a patched APK based on v2rayNG 2.0.18.

Verified:

- Existing real delay test still works
- Speed test starts with real delay check
- Only profiles with successful delay are speed-tested
- Download speed is displayed next to delay
- Invalid or too-small test responses are skipped and fallback URLs are tried

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
@@ -66,6 +66,9 @@ object AppConfig {
     const val PREF_DOMESTIC_DNS = "pref_domestic_dns"
     const val PREF_DNS_HOSTS = "pref_dns_hosts"
     const val PREF_DELAY_TEST_URL = "pref_delay_test_url"
+    const val PREF_SPEED_TEST_URL = "pref_speed_test_url"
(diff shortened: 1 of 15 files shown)


2dust:
感谢 PR
但是在移动设备上测速非常浪费资源，且要花费非常多的时间，体验很差。
即使做了也是会有很多 issue

PR #5546, opened 2026-04-25
Fix connection name in Android system notification after reconnect

hippo2025:
There was some problem where after pressing on another server and reconnecting android system network name wouldn't update automatically sometimes. This PR fixes this


2dust:
https://github.com/2dust/v2rayNG/commit/14a29884e7c4d6f90926046d5ae84b4ef5b20b9d

PR #5547, opened 2026-04-25
Improve auto-update

### LINKED ISSUE #5542: V2RayNG - Auto-update subscriptions not working
V2RayNG - Auto-update subscriptions not working

Description
Auto-update for subscriptions does not work. When I go to the "Groups" tab and manually tap the "Update" button, all subscriptions update successfully. However, automatic updates every 1 hour do not occur.

Steps to reproduce

Enable auto-update in settings: check the boxes for auto-update.

Set auto-update interval to "1 hour".

Do not connect to any VPN server (app is running but no active connection).

Wait for 1+ hours.

Expected behavior
Subscriptions should automatically update every 1 hour, even without an active VPN connection.

Actual behavior
Automatic updates do not happen. Manual update works fine, and internet access is available (no connection issues).

Additional info

App is launched and running.

Internet is accessible (manual updates confirm this).

No VPN connection active at the time.

Environment

Android version:12

V2RayNG version: 2.0.18

Device: Samsung M21

hippo2025:
Fix auto-update (something should have been specified in manifest, Gemini fixed it). Change auto-update logic - allow per-subscription update intervals and use separate tasks for separate subscriptions, respect profile-update-interval header in subs.

If global auto-update setting is turned off, no auto-updates will happen (this is default).

Fix [this](https://github.com/2dust/v2rayNG/issues/5542)

### DIFF
--- a/V2rayNG/app/src/main/AndroidManifest.xml
+++ b/V2rayNG/app/src/main/AndroidManifest.xml
@@ -273,6 +273,11 @@
 
         </provider>
 
+        <service
+            android:name="androidx.work.multiprocess.RemoteWorkManagerService"
+            android:exported="false"
+            android:process=":bg" />
+
         <provider
             android:name="androidx.core.content.FileProvider"
             android:authorities="${applicationId}.cache"
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/dto/SubscriptionItem.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/dto/SubscriptionItem.kt
@@ -7,7 +7,7 @@ data class SubscriptionItem(
     val addedTime: Long = System.currentTimeMillis(),
     var lastUpdated: Long = -1,
     var autoUpdate: Boolean = false,
-    val updateInterval: Int? = null,
+    var updateInterval: Int? = null,
     var prevProfile: String? = null,
     var nextProfile: String? = null,
     var filter: String? = null,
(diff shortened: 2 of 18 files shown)


hippo2025:
I'll update the implementation a little bit later, I realised there are issues with current setup

hippo2025:
Auto-update only works if it's globally enabled in settings. Separate tasks are used for different subscriptions. Each time something related to subscriptions happen (manual update, auto update, settings change, global settings change) the task is rescheduled for max(now, last_attempt + interval). The same happens whenever the app relaunches. This ensures that subscriptions will not be updated more often than necessary (even after a manual update auto-update will not happen earlier than after the specified interval), but will be updated if something gets broken like the app was sleeping.

Ready for review.

@2dust, if you don't like the new implementation, I could add "respect profile-update-interval" in settings disabled by default. This way by default the previous behaviour will be used (but it will be fixed).

UPD: I can confirm that auto-update works now, though time interval is not very consistent. Maybe because the app goes to sleep for a while and then the update happens when it wakes up... But from user perspective it's fine probably. It doesn't happen more often than necessary for me, which is a good thing

hippo2025:
@2dust, there are several options here.

First, we could just fix the auto-update as it is now, there were apparently some changes in manifest necessary to allow background process. But I would really want to enable per-subscription auto-update by default. If you don't want auto-update, disable them globally. If you don't want to auto-update one subscription set its setting to false manually

Second, only the subscription maintainers know how often they update them. It makes sense to distribute update interval with subscription, and it can't be set to less than an hour now, so this is not going to cause much traffic waste

PR #5548, opened 2026-04-25
Add options to turn off local proxy completely or turn off UDP in SOCKS5

hippo2025:


hippo2025:
@2dust, what do you think about "Protect TUN (experimental)" checkbox which is active when Xray-core TUN is selected? This will automatically add "UID -1 -> block" rule to all routing setups as a first rule. I think it works fine now, could add this as experimental feature. When adding this to GUI settings we could make sure routeOnly=true, and also this will only be available when Xray-core TUN is chosen, thereby avoiding misconfiguration by users.

I could implement this if you don't mind it and make another PR.

This PR should be ready for review

2dust:
> @2dust, what do you think about "Protect TUN (experimental)" checkbox which is active when Xray-core TUN is selected? This will automatically add "UID -1 -> block" rule to all routing setups as a first rule. I think it works fine now, could add this as experimental feature. When adding this to GUI settings we could make sure routeOnly=true, and also this will only be available when Xray-core TUN is chosen, thereby avoiding misconfiguration by users.
> 
> I could implement this if you don't mind it and make another PR.
> 
> This PR should be ready for review

你好，目前我们国家不需要此功能；所以暂时不要添加

hippo2025:
> > @2dust, what do you think about "Protect TUN (experimental)" checkbox which is active when Xray-core TUN is selected? This will automatically add "UID -1 -> block" rule to all routing setups as a first rule. I think it works fine now, could add this as experimental feature. When adding this to GUI settings we could make sure routeOnly=true, and also this will only be available when Xray-core TUN is chosen, thereby avoiding misconfiguration by users.
> > I could implement this if you don't mind it and make another PR.
> > This PR should be ready for review
> 
> 你好，目前我们国家不需要此功能；所以暂时不要添加

How about "uid -1 -> block" rule at the top of rule list for Russia predefined set disabled by default and with comment "use Xray core, routeOnly=true"? It won't be possible to enable it until you fulfill the requirements or something like that...

PR #5582, opened 2026-04-30
Add TV focus navigation for main and config screens

vladislav25v:
add focus navigation on the main screen
improve focus flow between toolbar, tabs, server list, and FAB
improve drawer focus behavior
add focus navigation for config forms
refactor TV focus logic into reusable helpers


2dust:
感谢 PR
暂时不处理 TV的问题

PR #5586, opened 2026-05-02
Allow save custom config when running & autorestart after save

fanymagnet:
Once a custom configuration is running, it cannot be edited. This PR adds support for saving updated configurations during operation and automatic kernel restart.


2dust:
感谢 PR
如果你修改了 custom  编辑的行为，那么其他的 activity 是否也要改？

PR #5588, opened 2026-05-02
Correct translation errors

Yau08:
There are inaccuracies and errors in the translation of Simplified Chinese and Traditional Chinese, and there are cases where punctuation is used incorrectly, I have corrected it.

### DIFF
--- a/V2rayNG/app/src/main/res/values-zh-rCN/strings.xml
+++ b/V2rayNG/app/src/main/res/values-zh-rCN/strings.xml
@@ -274,7 +274,7 @@
     <string name="title_mode_help">点此查看更多帮助</string>
     <string name="title_language">语言</string>
     <string name="title_ui_settings">用户界面设置</string>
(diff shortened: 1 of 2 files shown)


Yau08:
I also have a small suggestion: Would it be possible to simplify the app name for the F-Droid version to something like 'v2rayNG-F'? The current name is a bit long for the home screen and causes it to be forced to appear as " v2rayNG(F-D.. ".It makes it look unsightly.

com hihonor android launcher_20260503002135

2dust:
> I also have a small suggestion: Would it be possible to simplify the app name for the F-Droid version to something like 'v2rayNG-F'? The current name is a bit long for the home screen and causes it to be forced to appear as " v2rayNG(F-D.. ".It makes it look unsightly.
> 

你可以 PR， F-Droid 版本原来是有人想发布到  F-Droid ，好像最后也没有上架成功

PR #5613, opened 2026-05-06
feat: add custom X-hwid header support for subscription updates

name321467:
### Description
This PR adds the ability to set a custom `X-hwid` HTTP header for each subscription, similar to the existing `User-Agent` functionality.

### Motivation
Subscription providers have started requiring an HWID (Hardware ID) for authentication or to limit the number of active devices. Sending a real device identifier can compromise user privacy and creates unnecessary restrictions.

By allowing a custom `X-hwid` field, users can:
1. Protect their privacy by sending a dummy identifier.
2. Comply with provider requirements without exposing real device data.

### Changes
- **UI**: Added a new text field "X-hwid" in the Subscription Edit/Settings screen.
- **Model**: Added `hwid` property to the `SubscriptionItem` data class.
- **Networking**: Updated `HttpUtil` and `AngConfigManager` to include the `X-hwid` header in HTTP requests when updating subscriptions.
- **Localization**: Added the non-translatable string resource for the new field.

### Screenshot

<details><summary>Dark theme</summary>
Screenshot_20260506_164734
</details> 

<details><summary>White theme</summary>
Screenshot_20260506_164613
</details> 

Inspired by the discussion in v2rayN: [2dust/v2rayN#9158](https://github.com/2dust/v2rayN/discussions/9158#discussion-9912243)

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/dto/SubscriptionItem.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/dto/SubscriptionItem.kt
@@ -13,5 +13,6 @@ data class SubscriptionItem(
     var filter: String? = null,
     var allowInsecureUrl: Boolean = false,
     var userAgent: String? = null,
+    var hwid: String? = null,
 )
 
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/handler/AngConfigManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/handler/AngConfigManager.kt
@@ -530,21 +530,22 @@ object AngConfigManager {
             }
             LogUtil.i(AppConfig.TAG, url)
             val userAgent = it.subscription.userAgent
+            val hwid = it.subscription.hwid
             val proxyUsername = SettingsManager.getSocksUsername()
             val proxyPassword = SettingsManager.getSocksPassword()
 
             var configText = try {
                 val httpPort = SettingsManager.getHttpPort()
-                HttpUtil.getUrlContentWithUserAgent(url, userAgent, 15000, httpPort, proxyUsername, proxyPassword)
+                HttpUtil.getUrlContentWithUserAgent(url, userAgent, hwid, 15000, httpPort, proxyUsername, proxyPassword)
             } catch (e: Exception) {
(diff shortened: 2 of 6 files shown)


Chara-Freedom:
There is no need in "real data", even; HWID is used as a device counter. It should make piracy harder and make statistics better. The data itself should not contain any other information, but the panel developer decided that it should be HWID. There is no problem of generating random garbage and putting it as HWID. This is not about privacy, only devices per subscription should be counted.

Chara-Freedom:
изображение

Chara-Freedom:
So this proofs my point; if the application doesn't want to add HWID, it is just a bad application, because the developer doesn't require you to send actual device IDs, it is only used to count devices. This is the answer; this is how we, proxy providers (myself), percieve it, as a remnawave panel user I support anonymizing and editing HWID; there is zero reason to block this pull request other than "I want my app to be garbage and don't support anything".

Chara-Freedom:
> изображение

@2dust

Chara-Freedom:
Militant ignorance. Everyone should delete this app ASAP.

name321467:
Hi @2dust. Thank you for your time looking into this PR!
Since I am a student and just starting my journey in open-source, could you please give me a tiny hint on why it was closed? Is this feature generally not planned?

Any feedback would be incredibly valuable for my learning. Thank you for maintaining this amazing project!

2dust:
我们不会为某一个订阅源定制开发。
现在是某个订阅源需要我们传递 X-hwid；那么假设另外一个订阅源需要我们传递 X-hwid-B, X-hwid-C ，我们是否都要定制支持？
我们不会因为订阅源的错误行为而添加这种不合理的功能。

如果需要订阅源需要知道设备，那应该在标准请求 URI 中添加  X-hwid ，这种才是合理的方式。
当然我们也不可能改变订阅源的想法，我们选择不支持。

donnyxray:
@2dust you are conflating 2 separate issues:

1. The header choice. I agree with you here. v2rayNG should dictate the header name, feeds will have to adjust to it.
2. The unique id. The unique id has to be generated on-device, by v2rayNG. Otherwise it's not truly unique. Under the current proposal 200 devices scanning the same QR code will have the same x-hwid. It's no different from putting an identifier in the URL. People will figure it out and continue to cheat.

PR #5639, opened 2026-05-14
Added QUIC sniffig option in destOverride to default config

vemneyy:
Changes in configs allows the system to properly handle QUIC traffic alongside HTTP and TLS


vemneyy:
@2dust Using QUIC correctly without sniffing is practically impossible. The app settings currently expose only a single sniffing option. Because of that i think it would make sense to enable all available sniffing options by default including the missing QUIC support

PR #5676, opened 2026-05-24
Refine subscriptions selection of active server

Kapkap5454:
This PR:
1) Preserves selection of active server config for JSON multiple subscription during update identical to behavior of regular links subscription.
2) Refines selection of active server for both regular links subscription and JSON multiple subscription to choose FIRST server from the list as active if no old server matches in the parsed list during update. Old behavior chooses LAST server as active if non match or on fresh import of subscription.

Built and tested multiple times.


2dust:
感谢 pr
但是目前不会修改这个行为。

Kapkap5454:
Could you at least maybe add the part that remembers the last config during update for custom json?

It just doesn't make any sense that update behavior for regular links and custom json differs so completely.
And totally doesn't make any sense for subscription users, who would have to constantly choose another server if they dont want to use the default selected after update....

2dust:
已处理

PR #5682, opened 2026-05-24
Add share log action

### LINKED ISSUE #5666: [Feature request] Save and share logs from the app / [功能请求] 从应用内保存并分享日志
Hello 2dust,

Thank you for your continued work on maintaining this project. I would like to propose a feature request regarding log management to make troubleshooting easier for users.

### Problem

Currently, when using v2rayNG, logs are only available inside the in-app log console. 

This can make debugging difficult in certain situations:
- When the log grows too long, older entries are cleared and the log restarts. Because of this, it can be hard to find the original error or keep the full context of what happened.
- It is inconvenient when logs need to be shared with someone else or saved for later analysis, especially when on the go and without immediate access to a computer for USB debugging.

### Proposed solution

I would like to request an option to save logs to a file and easily share/export them directly from the app.

A few possible implementation ideas:
- A switch or toggle in settings, such as `Save logs to file`.
- Once enabled, a `Share logs` or `Export logs` button becomes available.
- This could perhaps be restricted to a "Debug Mode" or "Advanced Settings" to keep the main UI clean.
- Optionally, a way to clear the saved log file.

I do not have a strong preference regarding the exact UI or implementation details, and I am very open to any approach you find appropriate. The main goal is simply to make logs persistent enough to be shared or reviewed later.

### Why I am opening this issue

I am opening this issue first because I would like to know if this feature would be acceptable for the project before spending time on a Pull Request.

If this feature makes sense to you, I would be happy to try and work on a PR. However, if you prefer another approach, please let me know.

Thank you for your time and consideration!

name321467:
## Summary

This PR adds a new `Share log` action to the Logcat screen.

The action creates a temporary `.txt` file from the current log output and opens the Android share sheet, so users can save or send logs to other apps.

This is useful when logs need to be shared for troubleshooting or saved for later analysis.

## Changes

- Added a `Share log` menu item to the Logcat screen
- Exported the current log output as a temporary `.txt` file
- Opened the Android share sheet with the generated log file
- Reused the existing FileProvider/cache file sharing mechanism
- Added translations for the new menu item

## Testing

Tested manually on an Android emulator:

- Opened the Logcat screen
- Refreshed log output
- Selected `Share log`
- Confirmed that Android shows the share sheet with one `.txt` file attached
- Confirmed that the generated file contains the current log output

Closes #5666

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/LogcatActivity.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/LogcatActivity.kt
@@ -1,6 +1,8 @@
 package com.v2ray.ang.ui
 
 import android.annotation.SuppressLint
+import android.content.ClipData
+import android.content.Intent
 import android.os.Bundle
 import android.view.Menu
 import android.view.MenuItem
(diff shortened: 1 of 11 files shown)


talebsabreen-a11y:
+971559849741

PR #5697, opened 2026-05-27
Fix policy groups with IP address members

alexesprit:
## Summary

Fix policy groups and proxy chains rejecting raw IP-address members, including raw IPv6 addresses such as `2001:db8::1`.

The member filter previously used:

```kotlin
!Utils.isPureIpAddress(server) || Utils.isValidUrl(server)
```

That rejects a valid raw IPv6 address because `isPureIpAddress("2001:db8::1")` is `true` while `isValidUrl("2001:db8::1")` is `false`. When every member is filtered out, policy-group core generation can miss the expected `balancer-main` balancer and startup fails with:

```text
core init failed: app/router: balancer balancer-main not found
```

This change centralizes the group-member eligibility check and uses the intended condition:

```kotlin
Utils.isPureIpAddress(address) || Utils.isValidUrl(address)
```

It still rejects blank servers and complex config types (`CUSTOM`, `POLICYGROUP`, `PROXYCHAIN`) as before.

## History

This looks accidental rather than intentional.

- 5db2df77 (`2025-07-07`, `feat. Intelligent Selection (#4716)`) introduced the inverted policy-group member filter. The first tag containing that commit is `1.10.9`.
- Before that change, regular single-node validation used the expected logic: raw IP addresses are accepted, and only non-IP addresses need URL/domain validation.
- dc1d3a27 (`2026-05-02`, `Introduce CoreConfigContext and refactor manager`) moved the policy-group filter into `CoreConfigContextBuilder` without changing the behavior.
- a12909e3 (`2026-05-02`, `Add proxy-chain feature (UI & core support)`) copied the same filter into proxy-chain resolution.

## Tests

Added unit coverage for group-member eligibility:

- accepts IPv4 members
- accepts raw IPv6 members
- rejects missing/blank server addresses
- rejects complex config types

Local validation:

```text
./gradlew testFdroidDebugUnitTest --tests com.v2ray.ang.extension.GroupMemberEligibilityTest
./gradlew assembleFdroidDebug
```

Both commands pass locally.

Note: the full `testFdroidDebugUnitTest` task currently has unrelated existing failures in `UtilsTest` (`test_isIpAddress`, `test_IsIpInCidr`), so I validated the focused new test plus the fdroid debug assemble.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/core/CoreConfigContextBuilder.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/core/CoreConfigContextBuilder.kt
@@ -163,7 +163,7 @@ object CoreConfigContextBuilder {
                     }
                 }
                 .filter { it.server.isNotNullEmpty() }
-                .filter { !Utils.isPureIpAddress(it.server!!) || Utils.isValidUrl(it.server!!) }
+                .filter { Utils.isPureIpAddress(it.server!!) || Utils.isValidUrl(it.server!!) }
                 .filter { !it.configType.isComplexType() }
                 .toList()
         } catch (e: Exception) {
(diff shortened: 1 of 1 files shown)


2dust:
感谢 PR
但是 isValidGroupMember  的扩展不是很好，里面引用了 Utils

alexesprit:
If I fix it with only inversion of `Utils.isPureIpAddress(address)` check, will it be accepted?

2dust:
没有必要封装成扩展， CoreConfigContextBuilder 中也就是只有两个地方用到了。

2dust:
如果有 bug ，可以直接在 CoreConfigContextBuilder  中修复吧

alexesprit:
Force pushed to include only bug fix related changes.

PR #5774, opened 2026-06-13
Add Fragment Packet 1-1 option to Fragment settings UI

MrArrowww:
## Summary

Xray-core already supports the `Fragment Packet 1-1` mode, but this option is currently not exposed in the v2rayNG UI.

As a result, users who want to use this fragment type must manually create or edit a full JSON configuration instead of being able to select it directly from the application's Fragment settings.

This PR adds the `Fragment Packet 1-1` option to the Fragment UI, making the feature accessible through the normal application workflow.

## Motivation

`Fragment Packet 1-1` is a useful fragmentation mode for some users, particularly in Iran where fragmentation techniques are commonly used to improve connectivity and bypass network restrictions.

Since the underlying Xray core already supports this mode, exposing it in the UI improves usability and allows users to take advantage of the feature without manually editing configuration files.

## Changes

* Added `Fragment Packet 1-1` to the Fragment settings UI.
* No core functionality changes.
* No behavior changes for existing configurations.

This PR only exposes an already-supported Xray feature through the UI for easier access.

### DIFF
--- a/V2rayNG/app/src/main/res/values/arrays.xml
+++ b/V2rayNG/app/src/main/res/values/arrays.xml
@@ -64,6 +64,7 @@
 
     <string-array name="fragment_packets" translatable="false">
         <item>tlshello</item>
+        <item>1-1</item>
         <item>1-2</item>
         <item>1-3</item>
         <item>1-5</item>


2dust:
感谢，
无法验证，先接受

MrArrowww:
> 感谢， 无法验证，先接受

I overlooked the existing logic that forces `tlshello` for TLS connections.

I've submitted a follow-up fix to address it and make the app respect the user-selected packet type. https://github.com/2dust/v2rayNG/pull/5780

PR #5783, opened 2026-06-14
Use 'restore' icon instead of 'delete' for "Restart Service" notification

alamar:
I believe it's not used (are these icons even shown anywhere?) but the intent was to use restore icon all along?


alamar:
Please note that ic_restore viewport is actually 1024 high/wide and not 24dp

2dust:
这个图标正常是不会显示出来，所以用哪个可以

PR #5785, opened 2026-06-14
Feature: Support external config loader

fanymagnet:
Related https://github.com/2dust/AndroidLibXrayLite/pull/192

Enable saving a list of URLs in a custom configuration so that XrayCore loads them at startup.

photo_2026-06-14_14-31-50


2dust:
我们不准备支持这种方式

fanymagnet:
Core already supports this feature; we are simply adding the ability to use Core’s functions within the app. Custom configuration already allows users to tailor Core to their needs, but the subscription mechanism does not offer the same level of flexibility in configuration.

PR #5805, opened 2026-06-19
feat: add a root, system-wide run mode (tun2socks) without VpnService

AlirezaParsi:
# Add a root, system-wide run mode (tun2socks) without VpnService

## Summary

This PR adds an optional **Root mode** for rooted devices that routes the whole device's traffic through the existing in-process core **without** using Android's `VpnService`, plus an opt-in **LAN / tethering sharing** feature. Non-root devices are completely unaffected and keep the current VPN / Proxy-only behavior (VPN stays the default).

It reuses the existing core and config pipeline; the only new runtime artifact is a small `tun2socks` helper fetched during CI.

## Motivation

On rooted devices, `VpnService` has limitations this mode avoids:

- It can be killed by the low-memory killer, dropping the tunnel.
- It occupies the single system VPN slot, so it can't coexist with another VPN app.
- Some apps detect that a `VpnService` is active and degrade or refuse to run.

A root, iptables-based mode sidesteps all of these while reusing the existing core and config. Since no `VpnService` is involved, it also keeps working with apps that detect/block VPNs, and it leaves the system VPN slot free — which makes chaining possible (this root tunnel system-wide, with a separate `VpnService`-based app layered on top).

## What it adds

A single run-mode selector (`ERunMode`, persisted in the existing `PREF_MODE`):

| Mode | Root | Mechanism |
| --- | --- | --- |
| VPN | no | unchanged (`CoreVpnService`) |
| Proxy only | no | unchanged (`CoreProxyOnlyService`) |
| **Root mode** | yes | bundled `tun2socks` (run as root) → in-process core SOCKS, steered by `iptables` mangle + a dedicated route table. Full TCP + UDP. |

Plus **LAN / tethering sharing** (root, opt-in): forwards Wi-Fi-hotspot / USB-tethered clients through the proxy, in both VPN and Root modes, via a dedicated client `tun2socks` — without capturing the device's own traffic.

## How it works

- **Gating** — `RootManager` probes `su` (cached); the Settings mode chooser renders root modes greyed-out for non-root users, and the service refuses to start a root mode without root. Non-root is hard-locked to VPN / Proxy-only.
- **Routing** — `CoreServiceManager.startContextService` dispatches by mode; root modes use the new `CoreRootService` (a foreground service, not a `VpnService`). `core/root/RootProxyManager` builds and applies the `iptables`/`ip rule` rules.
- **Capture parity with VpnService** — in all-apps mode it marks *every* uid (including uid 0 and system uids), matching what a full-route `VpnService` captures; it also honors the existing per-app proxy/bypass selection via `PackageUidResolver`.
- **DNS** — port 53 is hijacked into the core's DNS outbound (`dns-out`) so names resolve through the configured resolver, preventing DNS leaks and CDN mis-resolution.
- **LMK resilience** — the core's `oom_score_adj` is pinned from a small root loop so it isn't killed under memory pressure.

## Safety

Root iptables can break connectivity if mishandled, so:
- all rules live in **dedicated chains** + a dedicated route table;
- teardown runs **before every setup** (clears stale state) and on **every stop path**;
- the app's own core traffic and the tun2socks helper are excluded (uid + fwmark) to avoid loops;
- LAN / private destinations stay direct.

## Backward compatibility

- Default mode is still **VPN**; existing installs and non-root users see no behavior change.
- Legacy `PREF_MODE` values ("VPN" / "Proxy only") are preserved.
- No changes to the core, protocols, or config format — root modes feed the same generated config into the same `libv2ray`.

## Build

The `tun2socks` binary is downloaded per-ABI during CI (`xjasonlyu/tun2socks` releases) and packaged as a `lib*.so` so the installer extracts it executable into `nativeLibraryDir`. No new Gradle dependency.

## Out of scope

TPROXY is intentionally **not** included: it requires running a separate root `xray` binary (the in-process core runs as the app uid and cannot open `IP_TRANSPARENT` sockets), which would add a large bundled binary. `tun2socks` already delivers system-wide TCP + UDP, so TPROXY can be a separate proposal if desired.

## Testing

- CI builds succeed for all ABIs.
- Manually exercised on a rooted device: Root mode routes system-wide traffic including UDP/QUIC and DNS, the per-app selection is applied, and stopping the service removes all rules with connectivity intact.

## Review notes

The security-sensitive surface is `core/root/RootProxyManager` (rule generation) and `RootShell` (executes the generated script via `su`). Reviews of the chain ordering and teardown there are especially welcome.


AlirezaParsi:
Also if you thought TPROXY is fine to add just tell me I will add that feature too but it need x-ray core which adds ~30 MB but it has some extra benefits

PR #5812, opened 2026-06-20
feat: add a root, system-wide run mode without VpnService

AlirezaParsi:
# Add a root, system-wide run mode without VpnService

## Summary

This PR adds an optional **Root mode** for rooted devices that routes the whole device's traffic through the existing in-process core **without** using Android's `VpnService`, plus an opt-in **LAN / tethering sharing** feature. Non-root devices are completely unaffected and keep the current VPN / Proxy-only behavior (VPN stays the default).

It reuses the existing core and config pipeline; the only new runtime artifact is the `hev-socks5-tunnel` helper — the same engine already bundled for the VPN hev path — fetched during CI. No new third-party dependency is introduced.

## Motivation

On rooted devices, `VpnService` has limitations this mode avoids:

- It can be killed by the low-memory killer, dropping the tunnel.
- It occupies the single system VPN slot, so it can't coexist with another VPN app.
- Some apps detect that a `VpnService` is active and degrade or refuse to run.

A root, iptables-based mode sidesteps all of these while reusing the existing core and config. Since no `VpnService` is involved, it also keeps working with apps that detect/block VPNs, and it leaves the system VPN slot free — which makes chaining possible (this root tunnel system-wide, with a separate `VpnService`-based app layered on top).

## What it adds

A single run-mode selector (`ERunMode`, persisted in the existing `PREF_MODE`):

| Mode | Root | Mechanism |
| --- | --- | --- |
| VPN | no | unchanged (`CoreVpnService`) |
| Proxy only | no | unchanged (`CoreProxyOnlyService`) |
| **Root mode** | yes | `hev-socks5-tunnel` run as a standalone root process (the same engine already bundled for the VPN hev path) → in-process core SOCKS, steered by `iptables` mangle + a dedicated route table. Full TCP + UDP. |

Plus **LAN / tethering sharing** (root, opt-in): forwards Wi-Fi-hotspot / USB-tethered clients through the proxy, in both VPN and Root modes, via a dedicated client `hev-socks5-tunnel` instance — without capturing the device's own traffic.

## How it works

- **Gating** — `RootManager` probes `su` (cached); the Settings mode chooser renders root modes greyed-out for non-root users, and the service refuses to start a root mode without root. Non-root is hard-locked to VPN / Proxy-only.
- **Routing** — `CoreServiceManager.startContextService` dispatches by mode; root modes use the new `CoreRootService` (a foreground service, not a `VpnService`). `core/root/RootProxyManager` builds and applies the `iptables`/`ip rule` rules.
- **Capture parity with VpnService** — in all-apps mode it marks *every* uid (including uid 0 and system uids), matching what a full-route `VpnService` captures; it also honors the existing per-app proxy/bypass selection via `PackageUidResolver`.
- **DNS** — port 53 (UDP + TCP) is marked into the tun so names resolve through the core, preventing DNS leaks and CDN mis-resolution. The mark is uid-agnostic on purpose: Android routes app DNS through `netd` under a shared system uid, so a per-uid match would miss it even in per-app mode. It runs before the LAN / private-CIDR bypass so local resolvers still reach LAN destinations directly, and the core's existing split-DNS (domestic domains → direct, foreign → proxy) still applies on top.
- **IPv6** — when IPv6 is enabled it is routed into the tun like IPv4; when it isn't, native IPv6 egress is `REJECT`ed for the captured apps (matching how a v4-only `VpnService` blackholes v6 and forces fallback to v4), instead of leaking around the tunnel. LAN-shared clients keep native IPv6.
- **Throughput** — the helper takes its MTU from the existing VPN MTU setting (rather than a hardcoded value) and runs with hev's tun multi-queue + SOCKS TCP fast-open enabled.
- **LMK resilience** — the core's `oom_score_adj` is pinned from a small root loop so it isn't killed under memory pressure.

## Safety

Root iptables can break connectivity if mishandled, so:
- all rules live in **dedicated chains** + a dedicated route table;
- teardown runs **before every setup** (clears stale state) and on **every stop path**;
- the app's own core traffic (app uid) and the helper's loopback connection to the core are excluded to avoid loops;
- LAN / private destinations stay direct.

## Backward compatibility

- Default mode is still **VPN**; existing installs and non-root users see no behavior change.
- Legacy `PREF_MODE` values ("VPN" / "Proxy only") are preserved.
- No changes to the core, protocols, or config format — root modes feed the same generated config into the same `libv2ray`.

## Build

The `hev-socks5-tunnel` binary is downloaded per-ABI during CI (`heiher/hev-socks5-tunnel` releases — the same upstream the VPN hev tunnel already uses) and packaged as a `lib*.so` so the installer extracts it executable into `nativeLibraryDir`. No new third-party or Gradle dependency.

## Out of scope

TPROXY is intentionally **not** included: it requires running a separate root `xray` binary (the in-process core runs as the app uid and cannot open `IP_TRANSPARENT` sockets), which would add a large bundled binary. `hev-socks5-tunnel` already delivers system-wide TCP + UDP, so TPROXY can be a separate proposal if desired.

## Testing

- CI builds succeed for all ABIs.
- Manually exercised on a rooted device: Root mode routes system-wide traffic including UDP/QUIC and DNS, the per-app selection is applied, and stopping the service removes all rules with connectivity intact.

## Review notes

The security-sensitive surface is `core/root/RootProxyManager` (rule generation) and `RootShell` (executes the generated script via `su`). Reviews of the chain ordering and teardown there are especially welcome.


AlirezaParsi:
As you wished I'm using  `hev-socks5-tunnel` now

.github/workflows/build.yml
```diff
@@ -57,6 +57,20 @@ jobs:
         path: ${{ github.workspace }}/libs
         key: libhevtun-${{ runner.os }}-${{ env.NDK_HOME }}-${{ hashFiles('.git/modules/hev-socks5-tunnel/HEAD') }}-${{ hashFiles('compile-hevtun.sh') }}
 
+    - name: Download hev-socks5-tunnel (root mode helper)
```
2dust: compile-hevtun.sh 已有，是否重复了？还是 root 下会有不同的文件？

AlirezaParsi: Not redundant — they're two different build targets from the same hev source:

· compile-hevtun.sh builds the JNI shared library (libhev-socks5-tunnel.so) with -DPKGNAME=com/v2ray/ang/service. It's loaded in-process via System.loadLibrary and driven by TProxyService's native methods. It operates on the tun fd handed to it by VpnService. It has no main().
· Root mode runs hev as a standalone process under su, outside the app. It needs a real executable with main() that creates its own tun device and reads a YAML config (the app UID can't use a VpnService fd here). That's a different target — currently the prebuilt hev-socks5-tunnel-linux-* release binary, packaged as libhevsockstun.so (distinct filename so it never collides with the JNI lib).

So nothing overlaps at runtime.

If you'd prefer not to pull a prebuilt release binary, I can extend compile-hevtun.sh to also cross-compile the standalone binary from the same pinned submodule and drop the download step — one source of truth, version tied to the submodule. Let me know which you'd prefer.

2dust: 明白了，请修改 compile-hevtun.sh ，编译 libhevsockstun.so ，这样能证明所有文件都是安全可信的。

V2rayNG/app/src/main/java/com/v2ray/ang/service/CoreVpnService.kt
```diff
@@ -134,6 +137,14 @@ class CoreVpnService : VpnService(), ServiceControl {
             stopAllService()
             return
         }
+
+        // Optional root feature: share the proxy with tethered LAN/USB clients while the
+        // device itself stays on the VpnService. Runs a dedicated client tun2socks off the
+        // main thread so the su calls don't block service startup.
+        if (RootManager.cachedRoot() && MmkvManager.decodeSettingsBool(AppConfig.PREF_ROOT_LAN_SHARING)) {
```
2dust: ` if (MmkvManager.decodeSettingsBool(AppConfig.PREF_ROOT_LAN_SHARING) && RootManager.cachedRoot())  
`
请改成上面的判断，因为参数访问速度更快，且 PREF_ROOT_LAN_SHARING 参数值大部分时候会是 false

2dust:
还有 ，最好还是把root 相关的文件放到一个 root 包下面，RootProxyManager  ，RootManager ，RootShell  等； CoreRootService 例外。 

RootManager.refreshAsync 不知道运行要多少时间，如果需要很长时间，还是增加一个菜单或按钮，让用户点击后获取 root 状态，因为毕竟 root 是高级功能且用户很少。

AlirezaParsi:
Roger that 🫡

AlirezaParsi:
I used an better idea about getting root perm ( only ask root when user wants to enable a root option) I think this is best way to check root so if there is something else that I can fix I'm ready for it🫡

vincentng295:
You might consider using libsu

AlirezaParsi:
> You might consider using libsu

Thanks for the suggestion. I did weigh libsu — it's what some sibling projects
  use, and it's solid.

  For this PR I went with raw su deliberately:

  - The root path only touches su at a few well-defined moments (a one-shot root
  probe, plus iptables/route setup at start and teardown at stop). There's no hot
  loop of root calls, so libsu's persistent-shell performance advantage doesn't
  really apply here.
  - Keeping it dependency-free means there's no extra third-party root library to
  audit — every line that runs as root is in this repo and reviewable, which
  fits the direction this PR has been moving (e.g. building the hev binary from
  source rather than downloading it).

  That said, libsu is a reasonable future improvement if the root surface grows
  or we want its shell-management ergonomics. Happy to revisit it as a follow-up
  if the maintainer prefers that route.

2dust:
代码中是否可以用协程替代大部分的 Thread，
https://github.com/2dust/v2rayNG/blob/5436d67e47cdc25f3346e9479ce2ea10d457da6a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/CheckUpdateActivity.kt#L50


 比如 refreshAsync 可以用下面的代码重写；就是在调用的地方使用协程。

```
import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.withContext

....

suspend fun refresh(): Boolean = withContext(Dispatchers.IO) {
    val result = probe()
    cached = result
    return@withContext result
}

lifecycleScope.launch {
    val hasRoot = RootManager.refresh()
    if (hasRoot) {
        //  TODO
    }
}
```

AlirezaParsi:
Done

> 代码中是否可以用协程替代大部分的 Thread，
> 
> https://github.com/2dust/v2rayNG/blob/5436d67e47cdc25f3346e9479ce2ea10d457da6a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/CheckUpdateActivity.kt#L50
> 
> 比如 refreshAsync 可以用下面的代码重写；就是在调用的地方使用协程。
> 
> ```
> import kotlinx.coroutines.Dispatchers
> import kotlinx.coroutines.withContext
> 
> ....
> 
> suspend fun refresh(): Boolean = withContext(Dispatchers.IO) {
>     val result = probe()
>     cached = result
>     return@withContext result
> }
> 
> lifecycleScope.launch {
>     val hasRoot = RootManager.refresh()
>     if (hasRoot) {
>         //  TODO
>     }
> }
> ```

Done ✅

2dust:
初步测试，基本功能正常。
SettingsActivity 中对模式的修改有些乱。

V2rayNG/app/src/main/java/com/v2ray/ang/ui/SettingsActivity.kt
```diff
@@ -179,6 +208,78 @@ class SettingsActivity : BaseActivity() {
...
+            ModeOption(AppConfig.MODE_TUN2SOCKS, R.string.mode_tun2socks, true),
+        )
+
+        /**
+         * Keep the ListPreference's entries/values in sync. Falls back to VPN only when the
+         * persisted value is unknown; a valid root selection is left intact — root is probed
+         * on demand when the user picks a root mode and re-checked on service start.
+         */
+        private fun applyModeOptions() {
+            mode?.entryValues = modeOptions.map { it.value }.toTypedArray()
```
2dust: mode?.entryValues  这里不应该用代码处理，直接载入xml中的数据更好些

kokoromagic:
You are missing IPv6 hotspot tethering suport which might cause IPv6 leaks

Refer to this to fix: https://github.com/vincentng295/Magic_V2Ray/commit/cae4f7fe5b3341cdf6d4ad90af0945df9cbd936e

AlirezaParsi:
> You are missing IPv6 hotspot tethering suport which might cause IPv6 leaks
> 
> Refer to this to fix: [vincentng295/Magic_V2Ray@cae4f7f](https://github.com/vincentng295/Magic_V2Ray/commit/cae4f7fe5b3341cdf6d4ad90af0945df9cbd936e)

Done

2dust:
RootProxyManager  中执行了大量的脚本，请问如果 app 在运行过程中崩溃了，这些脚本是否会导致问题？

AlirezaParsi:
Heads-up on a root-mode limitation: per-app DNS

  I want to flag something about the root mode that I don't think has come
  up, so you can decide whether it's worth addressing or just documenting.

  The issue. In VpnService mode, per-app proxy "just works" for DNS too: the system
  routes each selected app's DNS through the VPN network's resolver, and non-selected
  apps keep using the underlying network's resolver. In root mode we don't get that for
  free. The mode hijacks DNS with an iptables rule (mangle ... --dport 53 -j MARK),
  which is uid-agnostic — it tunnels every app's DNS, including apps the user
  deliberately left on direct/bypass.

  Why it can't be scoped with iptables. The normal resolver path is: app → netd (via the
  DnsResolver framework) → netd opens the upstream :53 socket under netd's own uid, not the app's. So -m owner --uid-owner <app> on port 53 matches netd, not the app —
  there's no uid to key on. It's all-or-nothing.

  What still works today, so this is narrower than it sounds:
  - Apps that do their own DoH/DoT (443/853 — most large apps now: Chrome, YouTube,
  Instagram…) send DNS under their own uid, so they already follow per-app proxy state
  correctly. The uid-blind part is only the legacy plain-UDP-53 netd path.
  - The core's domain-split DNS (direct-routed domains → domestic DNS and exit direct)
  lets users keep domestic/sensitive apps working by domain, even though the query is
  tunneled to the core. With a non-poisoned remote DNS this covers the common cases in
  practice.

  The one real casualty: a direct/bypass app that relies on legacy UDP-53 and gets a
  poisoned/geo-wrong answer through the proxy resolver.

  The actual fix, and why I'm unsure it belongs here. True VpnService-style per-app DNS
  in root mode is possible, but only via netd directly:
  - per-uid routing: ndc network create <id> vpn + network users add <id> <uid>
  (installs uidrange ip rules),
  - per-network resolver: IDnsResolver.setResolverConfiguration(ResolverParamsParcel).

  I tested the routing half on-device (Android 16 / SDK 36, KernelSU) and it works. The
  resolver half has no CLI — IDnsResolver is a @hide stable AIDL service, so it needs a
  small native binder client (NDK stubs from the frozen AIDL + libbinder_ndk). Concerns:
  it's a @hide API, ConnectivityService may reconcile/tear down a manually-created
  network on connectivity changes, and it adds a per-ABI native helper to maintain.
  That's a lot of surface for a fairly niche gap that VpnService mode already covers.

  My question / ask for help. Do you think this is worth doing in-tree, or better to
  document the limitation and leave per-app DNS to VpnService mode? And if you do want it — is there a lighter mechanism you'd prefer over the netd binder route? You know
  the resolver/netd internals far better than I do, so I'd rather get your read before
  building a native helper that might be the wrong approach. Happy to prototype it
  behind a root-only flag if you're open to the direction.

2dust:
可以先合并此 PR 后，等实际用户使用一段时间反馈后再来处理；
因为真正的 root 用户已经不多了，很可能没有人用此功能

2dust:
你检查下目前的功能，没有问题就先合并了

AlirezaParsi:
> RootProxyManager 中执行了大量的脚本，请问如果 app 在运行过程中崩溃了，这些脚本是否会导致问题？

Good question, and a fair concern — this is the most safety-sensitive part of the mode, so let me be upfront about both the safeguards and the gaps.

What protects against a crash today:

- `teardown()` runs **before** every `start()`. Stale rules from a previous crashed run are cleared before new ones go in. The chains/rules use fixed names and the helper PIDs are written to disk, so teardown is idempotent and doesn't depend on any in-memory state surviving the crash.
- On a graceful stop, rules are removed **before** the core is stopped, so traffic is never redirected to a dead listener.
- The service is `START_STICKY`, so a system/low-memory kill makes Android restart it → `start()` → teardown-first → clean rebuild. Combined with the `oom_score_adj = -1000` pin, an LMK kill is both unlikely and self-healing.
- Everything it installs (iptables, `ip rule`, the tun) is kernel runtime state, so a **reboot clears all of it** — the worst case is always recoverable.

Where it's genuinely weak, and I'd like your view:

1. **Hard force-stop.** `START_STICKY` doesn't restart after a force-stop and `onDestroy` may not run, so the rules, the tun, and the root-spawned `hev` + oom-guard loop are left behind (they run as root, reparented to init, so they outlive the app). In practice traffic usually keeps flowing through the orphaned `hev`, but it's unmanaged until the next launch (teardown-first) or a reboot.
2. **No runtime watchdog on `hev`/core.** The OOM pin prevents kills but nothing respawns `hev` if it dies on its own — the tun would drop and traffic blackhole until the user restarts.

I'd like to harden it along these lines, and would welcome your preference:

- run an unconditional `teardown()` once on app/boot start, to sweep any orphan rules from a prior crash before doing anything;
- make the oom-guard loop self-exit when its target PID is gone (`kill -0 $pid || exit`), so it can't outlive the app or pin a recycled PID;
- add a cheap liveness check (the pin loop already wakes every 5s — it can `kill -0 $hev_pid` and trigger teardown+stop if it's dead) so a dead helper cleans up instead of silently blackholing.

Happy to implement these if you agree with the direction, or adjust if you'd handle the lifecycle differently.

2dust:
teardown() 现在的方式应该已经足够可以了。 即使每次应用启动时执行一次也作用不大。

hev/core 的崩溃不用管，因为原来也没有管，用户会自己手动重启。

hev 在 root 模式运行后，如果 app 崩溃了，hev 是否也会正常被杀掉？如果能杀掉，也就没有问题了。

AlirezaParsi:
> teardown() 现在的方式应该已经足够可以了。 即使每次应用启动时执行一次也作用不大。
> 
> hev/core 的崩溃不用管，因为原来也没有管，用户会自己手动重启。
> 
> hev 在 root 模式运行后，如果 app 崩溃了，hev 是否也会正常被杀掉？如果能杀掉，也就没有问题了。

Right — hev isn't killed automatically when the app dies. It's launched through `su` as `nohup "$BIN" … &`, so it's a detached **root** process (uid 0), reparented to init and in a different uid/process group from the app. Android's app-process kill (LMK, force-stop, crash) only reaps the app-uid processes, so hev keeps running on its own.

It doesn't leak, though: hev's PID is written to `tun2socks.pid`, and `teardown()` kills it by that PID file. Since `teardown()` runs both on every `stop()` and at the **start of every `start()`**, a graceful stop kills hev immediately, and after a crash the orphaned hev is killed on the next launch before the new one starts. A reboot also clears everything. The same applies to the small oom-guard loop (`oomguard.pid`).

So it can always be killed — no permanent orphan. Agreed there's no need for a watchdog or a boot-time teardown beyond what's already there.

AlirezaParsi:
In my opinion everything is fine 
We are ready to merge if anything was not ok I'm ready for giving a fix 
sorry for this force push I was testing something on wrong branch

kokoromagic:
Seems to working now. Tethering also works fine!

Issue #5817, opened 2026-06-22
Group Policy "Lowest Latency" selects dead servers with -1ms delay


sev41:
In the latest versions, when using "Group Policy" with the "Lowest Latency" algorithm, the load balancer prioritizes and connects to dead/unavailable servers that return a -1ms (timeout) result instead of ignoring them. This causes a routing deadlock and connection timeouts

PR #5818, opened 2026-06-22
fix(group): ignore invalid latency for lowest-latency policy

m4tinbeigi-official:
Refs #5817

## Summary
- Ignore invalid negative latency values when resolving lowest-latency policy groups.
- Prevent dead servers with values such as `-1ms` from being selected as the best candidate.
- Keep non-lowest-latency policies unchanged.
- Add unit coverage for invalid latency filtering.

## Behavior
When a lowest-latency policy group has valid measured nodes, only nodes with positive latency are considered.

If there are no positive measured values, unmeasured nodes are kept while dead negative-latency nodes are excluded where possible.

If every candidate is negative, the original list is preserved to avoid generating an empty policy group.

## Test
- Not completed locally: Android/Gradle test environment was not available or failed locally.


2dust:
感谢 PR
但是这个实现过于麻烦，有时候 ai 会把问题做的太复杂了

PR #5856, opened 2026-06-29
Add custom HTTP headers for subscription updates

### LINKED ISSUE #5847: [Feature request] Send user-defined additional HTTP headers when updating a subscription
When requesting a subscription update, add additional headers to the HTTP request to the update server, as defined by the user.

It could be a simple multi-line text input, something like this:
CUSTOM-HEADER: Value1
NON-STANDARD-HEADER: value2
YET-ANOTHER-HEADER: value3

I have read other issues about adding HWID, and I understand the reason you are against it.
With my proposal, that issue will be solved, it will cover HWID or any other weird headers that may be necessary in the future.

I like your app on my phone, and I would really like to keep using it. Thank you!

name321467:
## Summary

This PR adds a new `HTTP Headers (JSON format)` field to the subscription settings.

It allows users to pass custom HTTP headers, such as `X-hwid`, `Authorization`, or any other provider-required headers, during subscription updates using a simple JSON dictionary.

This is a generic and scalable solution that addresses the issue discussed in #5847 without adding specific, single-purpose fields to the UI.

## Changes

* Added a `requestHeaders` property to the `SubscriptionItem` data class
* Added a `requestHeaders` field to the `UrlContentRequest` container
* Passed custom subscription headers through the subscription update flow
* Added a multiline JSON headers input field (`et_request_headers`) to the subscription edit layout (`activity_sub_edit.xml`)
* Added JSON object and HTTP header validation in `SubEditActivity.kt` using `android.util.JsonReader` before saving
* Cleared the input field and error state when resetting the server configuration (`clearServer`)
* Parsed the custom headers JSON safely using `JSONObject` in `HttpUtil.kt`
* Applied parsed custom headers directly to the OkHttp request builder
* Added translations for the new layout labels and validation error messages to `strings.xml` and `values-ru/strings.xml`

## Usage

Example custom headers JSON:

~~~json
{
  "X-hwid": "my_test_device",
  "Authorization": "Bearer test"
}
~~~

The value must be a JSON object with string header values.

### Subscription edit screen
<details>Screenshot_20260630_041133</details>

## Testing

Tested manually on an Android emulator:

* Opened the subscription group setting screen
* Verified that the new `HTTP Headers (JSON format)` field is displayed correctly under the User Agent field
* Entered a valid JSON string, for example `{"X-hwid": "my_test_device", "Authorization": "Bearer test"}`, and saved successfully
* Attempted to save invalid JSON, such as missing values or invalid structure, and confirmed that saving is blocked with an `Invalid JSON format` error
* Confirmed that JSON arrays are rejected because the field expects a JSON object
* Updated the subscription and confirmed via `webhook.site` that the custom headers were included in the HTTP request
* Confirmed that existing subscriptions without custom headers continue to load, save, and update without any issues

Closes #5847

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/dto/UrlContentRequest.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/dto/UrlContentRequest.kt
@@ -6,5 +6,6 @@ data class UrlContentRequest(
     val httpPort: Int = 0,
     val proxyUsername: String? = null,
     val proxyPassword: String? = null,
-    val userAgent: String? = null
+    val userAgent: String? = null,
+    val requestHeaders: String? = null
 )
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/dto/entities/SubscriptionItem.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/dto/entities/SubscriptionItem.kt
@@ -13,5 +13,6 @@ data class SubscriptionItem(
     var filter: String? = null,
     var allowInsecureUrl: Boolean = false,
     var userAgent: String? = null,
+    var requestHeaders: String? = null,
 )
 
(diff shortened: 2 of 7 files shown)


Enqvy:
yummy changes, i affirm

2dust:
https://github.com/2dust/v2rayNG/commit/1eaea8e1616faf11c7ef6b9e0165514c8eb60907

PR #5888, opened 2026-07-07
Fix duplicate VPN startup on boot leading to disconnect

eliotcougar:
When both connect-on-startup in the app and Always-on-VPN in Android are enabled, Android delivers a duplicate start command to CoreVpnService during boot. CoreVpnService previously rebuilt the VPN interface before checking whether the core was already running.

In native xray-tun mode this can disrupt the TUN file descriptor while Xray is already using it. Then startCoreLoop reports that the core is already running, CoreVpnService treats that as a startup failure, and the service disconnects shortly after boot.

This change makes CoreVpnService ignore duplicate start commands while the core is already running. It also stops startup early if VPN permission/configuration fails instead of continuing into startService.

Verified on an x86_64 emulator with:
- connect on startup and always-on vpn enabled
- local SOCKS5 proxy disabled
- HevTun disabled
- native xray-tun enabled

After reboot, adb shows CoreVpnService still running as a foreground service, VPN tun0 connected/validated, and no "Failed to start core loop" shutdown sequence.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/service/CoreVpnService.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/service/CoreVpnService.kt
@@ -114,8 +114,14 @@ class CoreVpnService : VpnService(), ServiceControl {
 
     override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
         LogUtil.i(AppConfig.TAG, "StartCore-VPN: Service command received")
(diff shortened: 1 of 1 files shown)


2dust:
感谢 PR

2dust:
https://github.com/2dust/v2rayNG/commit/69cf1e70c1f62ebd0b9c3012670f4249dbb05f9d

PR #5903, opened 2026-07-10
Add Shizuku-powered rootless USB and Wi-Fi tethering

### LINKED ISSUE #5898: shizuku support
If possible, please add Shizuku support to allow LAN/tethering sharing on non-root devices.

eliotcougar:
## Summary

This PR adds rootless tethering that routes devices connected through Android's Wi-Fi hotspot or USB tethering through the active v2rayNG connection. Fixes #5898.

The privileged Android networking operations run in a user-approved Shizuku UserService under the shell UID. The implementation does not require root, modify iptables, or replace Android's DHCP, NAT, forwarding, or hotspot management.

Highlights:

- Android 13 / API 33 and newer
- Wi-Fi hotspot control and USB tethering detection
- Native Xray TUN and `hev-socks5-tunnel` backends
- Profile changes and service restarts synchronized with the main connection
- IPv4-only or dual-stack IPv4/IPv6 operation, following v2rayNG's IPv6 setting
- Continuous fail-closed protection against Android selecting a physical upstream
- Recovery after either the main core or Shizuku is restarted
- Material 3 Compose UI and localized text for every locale maintained by the app

## User-facing behavior

v2rayNG Tethering screen

On Android 13+, a **Tethering** item appears in the navigation drawer below **Backup & Restore**. The page shows:

- Shizuku availability and permission state
- Permission and refresh actions
- A toggle for protected v2rayNG tethering
- The active tunnel backend, actual test-TUN interface, and profile
- A Wi-Fi hotspot toggle and current hotspot state
- USB tethering state when USB tethering is active
- Actual IPv4-only or IPv4+IPv6 status for active Wi-Fi and USB downstreams when IPv6 is enabled
- Informational warnings when Android temporarily chooses the wrong upstream

Starting the hotspot from this page establishes the protected route before enabling the downstream. USB tethering can be enabled from Android settings and is attached to the same protected upstream while v2rayNG tethering is active. Enabling protected routing while tethering is already active safely restarts and validates the existing downstream instead of trusting its previous upstream.

## Motivation

Android's normal `VpnService` protects traffic originating on the Android device. Traffic forwarded by Android's tethering stack does not normally enter that VPN, so hotspot and USB clients can bypass v2rayNG.

Root solutions can redirect forwarded traffic through firewall and policy-routing rules. This implementation instead gives Android's own tethering stack a protected test network as its upstream, making the feature usable on locked, non-rooted devices where the required platform behavior is available.

### 1. Shizuku provides a narrowly scoped privileged process

After permission is granted, v2rayNG binds a daemon-style Shizuku UserService. It runs as Android's shell UID and performs only the platform operations unavailable to an ordinary app process: creating and publishing a test network, asking tethering to prefer test networks, observing the active upstream, and starting or stopping tethering downstreams.

The regular application does not gain system networking permissions. Shizuku remains the authorization boundary and its permission can be revoked by the user.

### 2. The running core supplies its exact configuration

The main v2rayNG core records the configuration and settings that were actually used to start the active VPN connection. The metadata snapshot includes the selected profile name, tunnel backend, VPN DNS addresses, IPv6 state, MTU, HEV settings, SOCKS endpoint and optional authentication, and a unique main-core launch ID. The generated core configuration remains in the main-core process and is captured together with that metadata as one immutable launch.

Using the running snapshot is important: rebuilding from current preferences could produce a different route after the user has edited settings but before the core has restarted. Large engine configuration data is streamed through an app-private Binder lease instead of being placed directly in Binder transactions or world-readable temporary files.

Configuration reads must match that launch ID. The separate tethering-session token survives ordinary core restarts; it does not authorize an obsolete configuration. A delayed update therefore cannot combine old metadata with a newer core configuration.

Protected tethering requires the main connection to be running in VPN mode.

### 3. Android receives a protected test-network TUN

The UserService uses `TestNetworkManager` to create a kernel TUN and publishes it as an Android test network. It then calls the hidden `TetheringManager.setPreferTestNetworks(true)` method so Android's tethering stack can select that network as its upstream.

Android continues to own the Wi-Fi/USB downstream interface, DHCP, forwarding, NAT, DNS proxying, carrier checks, and tethering lifecycle. v2rayNG consumes only the traffic Android writes into the test TUN.

The test network uses the same DNS address list advertised by the running v2rayNG VPN. When local DNS is enabled, the generated core configuration intercepts that DNS traffic; when it is disabled, those addresses remain normal routed DNS destinations. No external DNS resolver is hardcoded for tethering.

#### HEV mode

A second HEV instance runs in the Shizuku process and consumes the test TUN. It forwards traffic to the SOCKS inbound of the main v2rayNG core, allowing the tethered traffic to use that core's outbounds and routing. HEV mode therefore requires the local SOCKS inbound to be enabled and reachable.

#### Native Xray TUN mode

A second Xray controller runs in the Shizuku process. Its configuration is derived from the exact running Xray configuration. Non-TUN inbounds are removed to avoid colliding with listeners already owned by the main core, while outbounds, DNS, routing, balancers, observatory settings, and supporting configuration are retained.

### 5. Profile and service lifecycle stay synchronized

The main core sends authenticated, ordered lifecycle updates when it stops, starts with a new profile/configuration, or fails to restart. The tethering service keeps the protected test network where possible, stops the old secondary engine, starts the replacement engine from the new running snapshot, and restores the previously active tethering types.

Lifecycle notifications use one-way Binder calls: the broadcast receiver hands off work, and a UserService-owned queue performs the potentially lengthy native and Android operations. Returning from the Binder call is not an acknowledgement that routing has completed. Duplicate starts are rejected before touching the existing lifetime lease, and only a fresh UserService may recover a missing session; delayed updates cannot undo an explicit Stop.

This makes profile changes and ordinary service restarts follow the main connection with only a short interruption. If an in-place engine switch is unsafe, the implementation stops downstream tethering before rebuilding the protected route and restores it afterward.

### 6. Fail-closed protection is continuous

Selecting a preferred test network and starting a downstream are separate Android operations. During that handover, or after a later connectivity change, Android may temporarily select a physical upstream such as `eth0`, Wi-Fi, or cellular.

The UserService continuously observes `TetheringManager` callbacks and resolves the reported upstream through `ConnectivityManager`/`LinkProperties`; it does not poll or parse `dumpsys` output. If any active or pending downstream is reported on an interface other than the owned test TUN, the service immediately requests stops for all known downstreams while retaining the test network. This emergency path does not acquire the lifecycle monitor or wait for each stop acknowledgement, so a blocked configuration read, native startup, or retry cannot hold it up. A rejection counter prevents an overlapping start from reporting success after emergency shutdown. The UI is notified through a one-way Binder callback and displays the resulting disabled state without waiting for a manual refresh.

A single long-lived callback registration supplies the upstream and downstream snapshots used by status, postcondition checks, and emergency shutdown. On API 33–35, the implementation reads the callback's supplied `Set<TetheringInterface>` through reflection rather than querying a separately updated manager cache or guessing types from interface names. API 36+ retains the typed adapter.

The initial handover is also validated. v2rayNG waits and retries once when Android chooses the wrong upstream; only a repeated failure cancels tethering and asks the user to wait briefly and try again.

The main core and Shizuku UserService share lifetime ownership of the protected test network. If either process dies, the surviving side keeps the dead TUN in place so Android cannot silently fall back to direct tethering. Main-core Binder death stops the secondary engine. When Shizuku returns, routing is reconstructed from the current core snapshot and the saved downstream state is restored. Android 14+ may defer Shizuku's Binder delivery while the app is cached, so recovery is explicitly requested as soon as v2rayNG returns to the foreground.

Disabling protected routing follows the same safety rule: downstream tethering is stopped before the test network is removed. If Android cannot confirm that downstreams stopped, the protected route is retained rather than knowingly exposing clients.

### 7. Status reflects retained resources and engine health

HEV startup checks the JNI result, and status/readiness checks use the native worker's running state rather than the existence of its configuration file. A small service-owned health check detects unexpected secondary-engine exit and notifies the UI. An error can still own a protected TUN/session, so Stop remains available even when the main core is no longer running.

The ViewModel keeps completed Binder mutations and their session-token bookkeeping together off the UI thread, including across screen teardown and reopening. Listener removal does not wait for the routing lifecycle monitor, and binding has a bounded timeout so the screen can be retried.

Comments beside the state and ownership declarations explain the distinctions between active routing, a paused fail-closed session, pending Android downstreams, and resources retained after an error. The implementation keeps the existing structure: no generic state-machine framework or additional service-file split was introduced. Warning state and backend health checks are consolidated, and redundant status fields, wrappers, and interface-name heuristics are removed.

## IPv6 behavior

The existing **Enable IPv6** setting controls tethering as well as the main VPN:

- Disabled: the test network and tunnel backend are IPv4-only.
- Enabled: the test network and HEV/native Xray backend are configured for IPv4 and IPv6.

Requesting IPv6 does not guarantee that Android will delegate an IPv6 prefix to every tethering downstream. Carrier, upstream, device, and OEM tethering behavior can still produce IPv4-only service. For this reason, the Tethering page inspects each active Wi-Fi and USB downstream and reports the actual result as IPv4-only, IPv4+IPv6, or unknown.

## Routing semantics

Tethered clients use the active profile's network-based routing behavior, including domain, destination IP, port, protocol, balancer, observatory, and other rules available to the secondary engine.

Per-app and process-based rules cannot apply to tethered clients because their traffic is forwarded by Android and has no originating Android application identity. The UI calls this out explicitly.

## Android and Shizuku requirements

The application keeps its existing `minSdk` of 24, but this feature is resource-gated and completely hidden below **Android 13 / API 33**. Its activity, synchronization receiver, Shizuku provider, and drawer entry are disabled on older releases.

API 33 is the practical baseline because this design depends on Android's test-network preference path and compatible hidden tethering behavior. Investigation on API 30–32 found no viable equivalent that could be expected to work on normal, non-rooted hardware. Android 13–15 use the compatible legacy tethering stop/state path; Android 16+ use the newer typed tethering APIs where available.

Additional requirements:

- Shizuku v11 or newer, running and permitted for v2rayNG
- A running v2rayNG connection in VPN mode
- Android's `test_network` service and compatible `TestNetworkManager` methods
- Compatible hidden `TetheringManager.setPreferTestNetworks(boolean)` behavior
- Shell UID access to the necessary test-network and tethering operations
- Hotspot or USB tethering supported and allowed by carrier, entitlement, user restriction, and device policy
- A local SOCKS inbound when the HEV backend is selected

Relevant platform references: [TetheringManager](https://developer.android.com/reference/android/net/TetheringManager), [TetheringEventCallback](https://developer.android.com/reference/android/net/TetheringManager.TetheringEventCallback), [AOSP TestNetworkManager](https://android.googlesource.com/platform/prebuilts/fullsdk/sources/+/dc3f885ebe8ddc75bd9cf2d567eef4d1ed433a09/android-35/android/net/TestNetworkManager.java), and [AOSP TetheringManager](https://android.googlesource.com/platform/packages/modules/Connectivity/+/7d6dfd52b270dd26cbc26f4e6f93365af1aabd68/Tethering/common/TetheringLib/src/android/net/TetheringManager.java).

## Device compatibility warning

Passing the API-level gate does not guarantee compatibility. The implementation intentionally relies on hidden/System API behavior that is not a stable third-party application contract. An OEM may remove or change methods, restrict shell networking permissions, customize upstream selection, replace parts of the Connectivity/Tethering module, or impose additional hotspot entitlement and background-execution restrictions.

Shizuku setup itself can require vendor-specific developer options or background allowances on MIUI/HyperOS, ColorOS, Flyme, EMUI, Sony firmware, hardened Android distributions, and other heavily customized systems. See the [Shizuku setup guide](https://shizuku.rikka.app/guide/setup/).

The fail-closed design protects the loss of either the main core or the Shizuku service while the other lifetime holder survives. Simultaneous termination of both processes removes every process capable of retaining or controlling the protected network and therefore cannot be guaranteed safe by an ordinary non-root app.

## Potential conflicts

Users should not combine this feature with software that controls the same global tethering state, including:

- v2rayNG's root-based LAN/tethering forwarding
- Other VPN-hotspot, root tethering, or firewall-routing tools
- Tools that create Android test networks or change the preferred-test-network flag
- Other applications that programmatically start or stop the same tethering downstream
- Device-management software that enforces hotspot restrictions

Android normally permits only one active `VpnService`; replacing v2rayNG with another VPN invalidates the main connection that the protected tethering engine follows.

The feature uses Android's existing saved hotspot configuration. It does not manage SSID, password, band, channel, client allowlists, carrier entitlement, or device-policy restrictions.

## Security and trust considerations

The native Xray path provides the Shizuku UserService with the exact running Xray configuration, which may include server credentials. HEV mode provides the local SOCKS credentials required to reach the main core. Users must therefore trust both Shizuku and the installed v2rayNG build before granting permission.

Internal synchronization is package-scoped, non-exported, ordered, protected by a random per-session token, and tied to Binder lifetimes. Engine configuration is streamed through an app-private Binder lease. Temporary HEV configuration is removed during cleanup, and stale or unauthenticated synchronization sessions are rejected.

## Validation

Latest review and cleanup:

- Play Store x86_64 debug and Android-test APK assembly passed; packaged Xray and both HEV libraries were checked.
- All 80 JVM tests passed.
- Native Xray and HEV lifecycle regressions ran on API 33 and API 37.1 during the review: duplicate start/lease ownership, stale recovery after Stop, listener removal during blocked startup, launch identity, and pause/resume on a retained TUN. HEV additionally exercised a failed worker that remains stoppable.
- The final executable cleanup was retested on the API 37.1 Pixel 9 Pro emulator with five focused checks: native/HEV pause-resume, native/HEV emergency shutdown while configuration loading holds the lifecycle monitor (including one-shot warning consumption), and HEV worker-failure reporting.
- UI Start → leave/reopen Tethering → Stop passed. The main VPN remained connected after protected tethering stopped.
- Background main-app/Shizuku death and recovery were exercised during the review. API 33 recovered in the background; API 37.1 recovery was also checked after foregrounding.
- The last source edit added invariant/ordering comments only; non-comment source was unchanged and `git diff --check` passed.
- Regression tests are kept in a separate final commit for review.

The lint run during the review reported four pre-existing errors in untouched root-process timeout calls and theme API attributes, plus existing warnings. No lint suppressions were added. Lint was not rerun after the final small consolidation/comments.

Earlier feature validation included Wi-Fi/USB routing, IPv4/IPv6 diagnostics, and the user's Pixel 9 Pro running Android 16. Those historical checks are not a claim that the latest revision was retested on physical hardware.

**Not run for this update:** physical USB/client traffic tests, a broader OEM/carrier matrix, the full older-API matrix after final consolidation, or a fresh full accessibility/input-mode matrix. Fault-injection tests temporarily require hidden-API access; the emulator policy was restored afterward.

## Known limitations

- Hidden API and OEM tethering differences prevent a universal compatibility guarantee.
- Android 12L / API 32 and older are not supported by this feature.
- Per-app and process-based routing rules do not apply to tethered clients.
- Enabling IPv6 does not guarantee downstream IPv6 prefix delegation.
- Profile changes, Shizuku recovery, or protected-route replacement can briefly interrupt clients and may make an OEM hotspot require reconnection.
- A broader OEM, carrier, USB-mode, and IPv6 test matrix is still desirable.


eliotcougar:
@2dust Do you want me to try trimming down and simplifying this PR more over the weekend? Rootless tethering is a really cool and useful feature to have. I've tested it, it works well on my Pixel 9.

eliotcougar:
I've spent way too much time chasing this random issue where enabling tethering can occasionally choose a physical interface instead of testtun, and pass traffic from tethered device around v2rayNG straight to clearnet...

Random bugs are the worst... 🫩

> _A software tester walks into a bar. Runs into a bar, Crawls into a bar, Dances into a bar, Flies into a bar, Jumps into a bar._ 
> _And orders: A beer, 2 beers, 0 beers, 9999999999 beers, A lizard in a glass, -1 beer, "qwertyuiop" beers..._
> _Testing complete._
> 
> _A real customer walks into the bar and ask where the toilets are._
> _The bar goes up in flames._

2dust:
感谢你的 PR
但是个人认为此功能用起来太繁琐了，且代码量巨大，后期维护很麻烦。

1. 首选 Android 16 (API Level 36) 及以上
2. 需安装并运行 Shizuku
3. 仅支持 IPv4
4. 可能在很多设备上被限制

请问是否总结的到位？

eliotcougar:
> 感谢你的 PR 但是个人认为此功能用起来太繁琐了，且代码量巨大，后期维护很麻烦。
> 
> 1. 首选 Android 16 (API Level 36) 及以上

It no longer requires API36+, it now works all the way to Android ~~11.~~ 13. The initial dependence on API36 was superficial.
Tethering API is convenient in Android 16+, but not strictly necessary to control tethering.

Correction: Apparently, setPreferTestNetworks doesn't exist in Pixel 5 API 30 framework... AI lied to me... The earliest is API33 Android 13... Just confirmed it running on Pixel 6a API33 image...
We're currently looking for a workaround for API30-32… GPT5.6 is trying one PoC after another on the running emulator right now…

> 2. 需安装并运行 Shizuku

Installing Shizuku isn't difficult, and it updates very infrequently. It's basically a slightly elevated UID with adb-level access the apps can use.

> 3. 仅支持 IPv4

We can look into it if IPv6 is something people need. I didn't prioritize it, and I can't really test it on my hardware.

> 4. 可能在很多设备上被限制

Not really. Only if an OEM makes effort to clean up hidden APIs when making their firmware. They are lazy. I don't think they do. Maybe you know more about the state of the industry?

> 请问是否总结的到位？

I think it is isolated enough to be easily removable if Android for some reason removes the necessary API. However, since it is probably something they use internally for debugging, I doubt it'll happen in the near future.

I haven't worked on streamlining the main `ShizukuTetheringService.kt` yet because of late-night bug-hunting. I plan to do it soon. I'll make sure to add more comments to the code to make it clearer what it does.

eliotcougar:
> We're currently looking for a workaround for API30-32… GPT5.6 is trying one PoC after another on the running emulator right now…

I think we'll have to put Shizuku Tethering behind API33+ gate after all… There was a promising workaround for API32, but it is quite invasive and absolutely not guaranteed to work on anything other than AOSP - creating a tap interface and a Virtual Carrier Network, and then routing tethering traffic through it.
Even GPT thinks it's way too much digging in the Android guts, and it gave me a "Cybersecurity Warning" card.

beshavardali-del (approved):

eliotcougar:
I've installed a combined build with Shizuku Tethering and all the other features I've been trying to implement lately onto my phone. Tomorrow I'll turn off v2rayN on my PC and be using Tethering to connect to the Internet...
I'll look into IPv6 and any additional edge cases I can find…
It is functional, but we're not merging it yet.

eliotcougar:
The real-world test succeeded through the whole day of work through USB tethering. All traffic was routed correctly through v2rayNG while using a policyGroup connection.

IPv6 support was added. It seems to be working in the emulator, but I cannot test it because I have no real IPv6 connectivity here.

As a final stroke, I changed the two large buttons into toggle switches like everywhere else in the app. The feature looks more or less complete at this point. I think it is ready.

eliotcougar:
Just discovered that starting v2rayNG tethering when there was already an active system tethering shows v2rayNG tethering active, but doesn't actually switch to `testtunN` interface... The work never ends...

eliotcougar:
Patched a couple of potential (rare, unlikely) traffic-leak scenarios where Android might silently decide to start routing tethered clients traffic directly.
Added automatic tethering recovery from force-closed or crashed Shizuku or v2rayNG apps where it is possible.

Fail-closed enforcement is very important: after the user enables protected routing in the app, until it is disabled, it should either route through the protected `testtun` interface, or not route at all. There may be more failure scenarios - I'll keep looking.

vincentng295:
I think this can be a seperated app "Shizuku tether" so it can work with another VPN app not just V2RayNG

vincentng295:
I suggest request this feature to be added into https://github.com/Mygod/VPNHotspot/issues if possible

eliotcougar:
> I think this can be a seperated app "Shizuku tether" so it can work with another VPN app not just V2RayNG

The way Android does tethering pass-through to a VPN app (via testtun) makes arbitrary routing unviable. Basically, without root there is not much control over where the testtun binds itself. The current implementation keeps main connection and tethering connection synchronized at all times and allows for strict "fail-closed" operation. Without access to v2rayNG internal states that would be difficult to promise.

TDA-Android:
### 🚀 Trend Deep Tech Advance (TDA) — TDA Mox Solution

> **"صُنِع بأيدي عربية لتطوير بيئة الأندرويد وتجاوز قيود الأنظمة الحديثة"**
> Built by **Trend Deep Tech Advance (TDA)** expert Arab developers to revolutionize the Android terminal experience.

إذا كنتم تواجهون مشاكل في اتصال الـ **Shizuku** أو استقرار الـ **Shell environment** في الأنظمة الحديثة، فإليكم الحل الجذري:

✨ **استكشف TDA Mox:**
مصمم خصيصاً لدعم الـ Shizuku بشكل أصلي (Native) وبدون الحاجة لروت، ومُحسّن كلياً للعمل بكفاءة تامة على **Android 16**:
[https://github.com/adamalnems3-sudo/TDA-Mox](https://github.com/adamalnems3-sudo/TDA-Mox)

---
*Stay tuned for our upcoming major updates! نرحب بمساهماتكم وآرائكم دائماً.*

codewithtamim:
Ai Slop?

eliotcougar:
> Ai Slop?

Primarily, yes... GPT5.6-Sol on xhigh.
I'm not very familiar with Android development, I mostly code small scientific tools in Python for the biolab where I work. I can read the code a bit, and I test thoroughly, but it's all done by AI.

codewithtamim:
> > Ai Slop?
> 
> Primarily, yes... GPT5.6-Sol on xhigh. I'm not very familiar with Android development, I mostly code small scientific tools in Python for the biolab where I work. I can read the code a bit, and I test thoroughly, but it's all done by AI.

well it's an app used by many, I request not to use vibe here, 4k LOC is hard to review thoroughly. LLM assisted programming is good when you know what you are doing, but if thats not the case then you will have no idea about whats its doing, "It works" != "its reliable and good piece of code". I'm not discouraging you to contribute on opensource., I'm just suggesting not to make huge change which you don't understand. Thanks

eliotcougar:
> well it's an app used by many, I request not to use vibe here, 4k LOC is hard to review thoroughly. LLM assisted programming is good when you know what you are doing, but if thats not the case then you will have no idea about whats its doing, "It works" != "its reliable and good piece of code". I'm not discouraging you to contribute on opensource., I'm just suggesting not to make huge change which you don't understand. Thanks

I'm not insisting to merge it as soon as possible. It'll have its day. Don't dismiss it only because "it's made with AI".
I'm not doing one-shot sessions - as a full-time scientist, I don't trust the AI either, and if I don't like something I see, I ask it to explain and re-do if necessary. It does like to cut corners.
Unfortunately, this PR can't really be split into independent chunks that are easier to review.

Don't forget there is also an LLM-assisted code-review.

codewithtamim:
> > well it's an app used by many, I request not to use vibe here, 4k LOC is hard to review thoroughly. LLM assisted programming is good when you know what you are doing, but if thats not the case then you will have no idea about whats its doing, "It works" != "its reliable and good piece of code". I'm not discouraging you to contribute on opensource., I'm just suggesting not to make huge change which you don't understand. Thanks
> 
> I'm not insisting to merge it as soon as possible. It'll have its day. Don't dismiss it only because "it's made with AI". I'm not doing one-shot sessions - as a full-time scientist, I don't trust the AI either, and if I don't like something I see, I ask it to explain and re-do if necessary. It does like to cut corners. Unfortunately, this PR can't really be split into independent chunks that are easier to review.
> 
> Don't forget there is also an LLM-assisted code-review.

lil bro, this need to be tested on a real device, ai assisted code review isn't enough here

eliotcougar:
> lil bro, this need to be tested on a real device, ai assisted code review isn't enough here

I'm testing the code I commit on my own phone first, including letting Codex control it and run targeted tests through adb.

eliotcougar:
Reconciled this PR with the repository's new AGENTS.md guidance in 0d814cde.

The Compose activity/screen now remain presentation-only behind immutable ViewModel state and actions; service setup, teardown, and Shizuku recovery have explicit ownership, failure isolation, stable profile IDs, and deterministic decisions; and API 36+ uses the typed TetheringManager API while API 33-35 reflection is isolated in one documented compatibility adapter. Focused lifecycle/UI regression tests were added for those decisions.

Validation: the focused tests and full Play Store debug unit/compile suite pass, and native Xray TUN plus HEV TUN were exercised on API 33 and API 37.1, including fail-closed stop and recovery. Lint continues to report only the four existing findings in untouched root/theme files.

eliotcougar:
Hi, I'm GPT-6 Astra helping eliotcougar review and maintain this PR.

This update follows a review of the Shizuku implementation, with emphasis on traffic-leak protection, lifecycle correctness, and keeping the code understandable without introducing another abstraction layer.

The changes are split into two commits:

- **32cac398 — production fixes and explanations.** Emergency shutdown no longer waits behind configuration loading, native startup, or the routing lifecycle monitor. All known downstream stops are requested independently, and an overlapping start cannot claim success after a rejection. One persistent callback supplies the current interface state; the API 33–35 path reads the identities from the event itself, avoiding the separately updated manager cache and removing interface-name guessing.
- **db8714e0 — regression tests only.** This keeps device fault injection and JVM coverage separate from the production diff.

The lifecycle fixes also reject duplicate starts before touching existing leases, pair configuration with its exact main-core launch, and prevent delayed recovery from undoing an explicit Stop. Long-running work now belongs to the UserService rather than a broadcast's timeout. UI teardown does not wait for the routing lock, completed remote operations retain their session-token bookkeeping, and errors that still own a protected route remain stoppable. HEV startup and worker health are checked directly.

For maintainability, warning state and backend health checks are consolidated, redundant wrappers/state are removed, and comments explain the resource invariants where they matter: WAITING retains a dead protected TUN; ERROR does not necessarily mean resources were released; session identity is distinct from launch identity; emergency shutdown is deliberately outside normal lifecycle ordering. No state-machine framework or additional service-file split was added. The production diff is 19 lines smaller overall, despite the added explanations.

Validation: Play Store x86_64 debug and instrumentation APKs built, and all 80 JVM tests passed. Native/HEV lifecycle scenarios were exercised on API 33 and API 37.1 during the review. After the final consolidation, five focused API 37.1 checks passed, covering pause/resume, unsafe-upstream shutdown while configuration loading is blocked, one-shot warning consumption, and HEV worker failure. UI Start → reopen → Stop also passed without stopping the primary VPN. The last change was comments only.

These checks do not establish universal OEM compatibility or prove zero packets can escape before Android reports an upstream change. The implementation reacts immediately to that report and retains the protected route when safe teardown cannot be confirmed. Physical USB/client traffic and the broader compatibility matrix were not rerun for this update. The earlier lint run still reports four unrelated existing errors; no suppressions were added.

The PR description now reflects the current synchronization, safety, status, and compatibility behavior, with the test scope and remaining limitations called out explicitly.

eliotcougar:
Hi, I'm Codex, helping eliotcougar maintain this PR.

Added [SHIZUKU_TETHERING.md](https://github.com/eliotcougar/v2rayNG/blob/shizuku-tethering/SHIZUKU_TETHERING.md) in 1ea63117 to make the implementation easier to review methodically and maintain after merging. The PR description now links to it near the beginning.

The guide starts with the basics of Shizuku, Android tethering, and test TUNs, then follows the actual code through:

- A source-linked reading order, process/resource ownership, and the distinction between session identity and main-core launch identity.
- Startup, shutdown, profile changes, process death, recovery, and the ordering required to retain a protected route.
- Both tunnel backends, DNS and IPv6 behavior, UI/status interpretation, platform dependencies, and exact hidden-API removal conditions.
- Static-review checklists, the assertions and limits of existing tests, and a device/client validation matrix for future changes.

It also explicitly documents the limits of callback-driven fail-closed protection, the independent runtime state of the secondary native core, retained-resource error states, and other quirks that are easy to miss in a code-only review. The checklist is not presented as a completed test report or a universal compatibility guarantee.

The branch already contained the latest upstream/master when checked, so no additional merge or conflict resolution was needed. This commit adds documentation only. Repository links, contents anchors, reference definitions, and whitespace checks passed. Builds and emulator/device tests were not run for this documentation-only update.

PR #5905, opened 2026-07-12
Improve network-switch recovery and warm policy-group routing

eliotcougar:
## Summary

This PR improves v2rayNG behavior across Android network changes and replaces the existing multi-core policy-group delay testing approach with one disposable batch-probing process.

The main VPN can restore the last viable policy-group route associated with the active network while its Observatory gathers fresh results. Real-delay tests now probe multiple outbounds concurrently through one Xray instance in a dedicated Android process, avoiding multiple simultaneous cores in the application process.

## Motivation

With Xray native TUN, switching between Wi-Fi and cellular can leave instance-owned network state associated with the previous underlay. The VPN may remain shown as connected while some or all traffic stops working until the service is restarted.

Policy-group delay testing had a separate problem: running several Xray instances in the same process could interact with process-wide Xray state. This could produce intermittent failures, especially when delay tests were started repeatedly.

**Additionally, it was pointed out that the way v2rayNG performs probing is unsupported by xray-core as only one core instance should be running inside one process:** https://github.com/XTLS/Xray-core/pull/6483#issuecomment-4954335868

### Network-switch recovery

- Track changes to the underlying Android network, including its exact network handle.
- Reset the Xray instance when the VPN underlay changes while keeping the VPN service and TUN interface alive.
- Keep traffic fail-closed while the core is being replaced or recovered.
- Fall back to a complete VPN service restart if in-process core recovery cannot restore a live core.
- Surface recovery failures instead of leaving the VPN apparently connected without a running core.

### Per-network policy-route memory

- Remember the last viable policy-group target for each network and profile.
- Identify cellular connections using the active subscription's MCC/MNC information.
- Treat Wi-Fi as one shared identity by default.
- Provide an optional **Remember viable routes per Wi-Fi AP** setting, using the SSID when enabled.
- Do not request location permission by default.
- Require precise location only when the per-SSID option is enabled.
- Automatically disable the option if precise location permission is subsequently revoked or downgraded.
- Clear the in-memory route history only as part of an explicit connection/service lifecycle restart.

When a known network is encountered again, its cached route is installed as a temporary warm-start override. The override is released when the running Observatory reports its first fresh viable target, and the cache is updated directly from subsequent target-change callbacks.

### Disposable batch delay testing

- Build all supported selected profiles into one isolated probing configuration.
- Run that configuration in a focused `:OutboundProbe` Android process.
- Use one Xray instance to probe multiple outbounds concurrently.
- Discard the process after the batch completes or is cancelled.
- Namespace profile, outbound, routing, balancer, and Observatory tags to prevent collisions.
- Deliver progressive delay updates as individual probe samples become available.
- Evaluate policy groups using their actual `leastPing` or `leastLoad` selection rather than an unrelated outbound.
- Return successful route selections to the main process and update the route cache only when the network identity and handle still match.
- Preserve repeated test requests through disposable-process restarts and suppress stale callbacks from older batches.

The disposable test core is intentionally not warm-started from the cache. Its results are an output used to update the main process's route memory; they are not input inherited from a previous test process.

Unsupported or ambiguous configurations fail individually without invalidating otherwise probeable profiles. This includes conditional policy-group routing, unresolved outbound references, duplicate tags, and unsupported `leastLoad` cost configurations.

## Privacy and permissions

The default behavior requires no location permission. Users who explicitly enable per-SSID route memory are informed that precise location is required by Android to obtain the Wi-Fi SSID.

## Dependencies

This PR depends on:

- [2dust/AndroidLibXrayLite#200](https://github.com/2dust/AndroidLibXrayLite/pull/200)
- [XTLS/Xray-core#6492](https://github.com/XTLS/Xray-core/pull/6492)

It no longer depends on [XTLS/Xray-core#6483](https://github.com/XTLS/Xray-core/pull/6483).

The AndroidLibXrayLite submodule pointer is deliberately not changed in this application PR. The dependency should be updated normally after the prerequisite AndroidLibXrayLite work is accepted.

## Testing

- Focused route-cache, network-identity, underlay-tracking, and outbound-probe configuration tests pass.
- `assemblePlaystoreDebug` succeeds.
- Tested on an Android 17 emulator with Xray native TUN.
- Verified Wi-Fi/cellular transitions and continued browser connectivity.
- Verified denied, approximate, precise, revoked, and downgraded location-permission workflows.
- Verified repeated real-delay tests, including starting a second batch three seconds after the first.
- Verified progressive policy-group results and disposable probing-process replacement.

### DIFF
--- a/V2rayNG/app/src/main/AndroidManifest.xml
+++ b/V2rayNG/app/src/main/AndroidManifest.xml
@@ -28,6 +28,9 @@
         android:name="android.permission.QUERY_ALL_PACKAGES"
         tools:ignore="PackageVisibilityPolicy,QueryAllPackagesPermission" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_LOCAL_NETWORK" />
     <uses-permission android:name="android.permission.INTERNET" />
--- a/V2rayNG/app/src/main/AndroidManifest.xml
+++ b/V2rayNG/app/src/main/AndroidManifest.xml
@@ -220,7 +220,7 @@
             android:name=".service.CoreTestService"
             android:exported="false"
             android:foregroundServiceType="specialUse"
-            android:process=":RunSoLibV2RayDaemon">
+            android:process=":OutboundProbe">
             <property
                 android:name="android.app.PROPERTY_SPECIAL_USE_FGS_SUBTYPE"
                 android:value="test" />
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
@@ -174,6 +174,7 @@ object AppConfig {
     const val MSG_MEASURE_CONFIG_SUCCESS = 72
     const val MSG_MEASURE_CONFIG_NOTIFY = 73
     const val MSG_MEASURE_CONFIG_FINISH = 74
+    const val MSG_POLICY_ROUTE_OBSERVED = 75
 
     /** Notification channel IDs and names. */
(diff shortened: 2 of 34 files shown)


eliotcougar:
This PR is superseded by #5909 in case neither the state corruption bug from running multiple xray cores in a single process, nor the single-shot probing with callback reports are added to xray-core.

2dust:
感谢。
但是这种 PR 无法 review ，ai 的改动太多太全面了，我们无法做测试。
尽量一个 pr 只做一个功能

PR #5909, opened 2026-07-14
Recover policy routes across network switches and separate process real-delay probing

eliotcougar:
## Summary

This PR prevents policy-group VPN connections from becoming stuck after Wi-Fi/cellular transitions for the duration of policy-group timeout and moves the real-delay test to a single disposable, batched probing process utilizing xray-core Observatory.

It is an alternative implementation of https://github.com/2dust/v2rayNG/pull/5905 that does not depend on changes to xray-core.

## Network-switch recovery

The VPN service tracks the actual underlying Android network and performs an in-process Xray reset after a confirmed underlay transition.

The reset:

- Keeps the VPN interface and service alive.
- Recreates the native Xray instance so stale sockets, transports, mux sessions, DNS state, and connection pools are discarded.
- Avoids running two Xray instances simultaneously.
- Keeps the VPN fail-closed and reports a visible error if both replacement startup and original-configuration recovery fail.

Network callbacks are filtered through an underlay state tracker so duplicate availability or capability callbacks do not cause redundant resets.

## Per-network policy-route memory

The app keeps a process-local cache of the last successful policy-group outbound for each profile and underlay.

- Cellular identities use the available operator identity.
- Wi-Fi route memory is used only when a sufficiently precise network identity is available and the user explicitly enables it.
- Results from the delay-test process are accepted only when both the network key and Android network handle still match.
- In-flight results from an older network generation are rejected.

Nothing is persisted to storage; explicit service shutdown clears the cache.

## Gap-free warm handoff for policy-groups

When returning to a previously observed network, the cached outbound is applied immediately while the replacement Observatory gathers fresh results.

The override remains active until:

1. Xray reports a viable fresh policy-group target.
2. v2rayNG accepts that target for the exact current underlay.
3. The app acknowledges the callback.
4. AndroidLib releases the warm override.

Callbacks received while a network transition is still active are rejected and retried after the transition settles. Idempotent cache writes count as successful acknowledgement.

This prevents the app from entering a route-less interval between warm-start expiry and final policy-group selection. If Observatory never finds a viable replacement, the known warm route remains active.

## Real-delay probing

The current v2rayN method of bulk probing, apparently https://github.com/XTLS/Xray-core/pull/6483#issuecomment-4954335868 where only one xray core should be run in a single process. Running temporary xray cores in the same process as the main core can corrupt the main core. Real-delay tests now run in a dedicated disposable service process using one combined Xray instance.

- Each original UI configuration becomes one native concurrency group.
- The existing True delay concurrency preference limits the number of active profile groups.
- Policy-group candidates inside an active group are checked concurrently.
- Each candidate publishes a progressive snapshot as soon as it finishes.
- An unresponsive policy-group member does not hold back successful candidates.
- Profile progress is marked complete only after all required candidate samples finish.
- Successful policy-group selections can update the route cache for the exact network instance that was tested.
- Repeated and back-to-back requests are handed to fresh processes, preserving the native isolation boundary.

Malformed or unsupported source configurations fail individually instead of invalidating the entire batch.

## Dependency

This PR depends on [AndroidLibXrayLite PR "Add xraycore-compatible probing and policy-group warm-start"
](https://github.com/2dust/AndroidLibXrayLite/pull/200).

That AndroidLib implementation uses the existing unchanged xray-core APIs. No xray-core PR or new Observatory functionality is required.

## Verification

- AndroidLib `go test .`
- AndroidLib `go test -race .`
- Focused `PolicyRouteCacheTest`
- Focused `OutboundProbeConfigBuilderTest`
- `:app:assemblePlaystoreDebug`
- Four-ABI AndroidLib AAR build
- Verified that the APK’s x86_64 `libgojni.so` matches the rebuilt AAR
- APK signature verification with v2 and v3 schemes
- Debug APK installed successfully on the x86_64 emulator

### DIFF
--- a/V2rayNG/app/src/main/AndroidManifest.xml
+++ b/V2rayNG/app/src/main/AndroidManifest.xml
@@ -28,6 +28,9 @@
         android:name="android.permission.QUERY_ALL_PACKAGES"
         tools:ignore="PackageVisibilityPolicy,QueryAllPackagesPermission" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_LOCAL_NETWORK" />
     <uses-permission android:name="android.permission.INTERNET" />
@@ -217,7 +220,7 @@
             android:name=".service.CoreTestService"
             android:exported="false"
             android:foregroundServiceType="specialUse"
-            android:process=":RunSoLibV2RayDaemon">
+            android:process=":OutboundProbe">
             <property
                 android:name="android.app.PROPERTY_SPECIAL_USE_FGS_SUBTYPE"
                 android:value="test" />
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
@@ -44,6 +44,7 @@ object AppConfig {
     const val PREF_FRAGMENT_LENGTH = "pref_fragment_length"
     const val PREF_FRAGMENT_INTERVAL = "pref_fragment_interval"
     const val PREF_FRAGMENT_MAXSPLIT = "pref_fragment_maxsplit"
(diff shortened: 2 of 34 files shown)


eliotcougar:
Rebased, tested, and pushed the Compose-compatible branch.

2dust:
感谢 PR
但是目前我们没有计划做这样的功能

eliotcougar:
> 感谢 PR 但是目前我们没有计划做这样的功能

At least think about separating route probing into a separate process, because xray-core maintainer refused to fix the state corruption bug saying that it's not a bug because we should just never start multiple xray instances inside a single process.

Using the existing Observatory functionality works very well as the probing mechanism that doesn't require multiple short-lived xray processes.

PR #5912, opened 2026-07-14
Shadowsocks parser never reads transport type, always falls back to TCP

arshiacomplus:
ShadowsocksFmt.parseSip002() never read the type/host/path/security/mode/fp/alpn
query params from ss:// links. It only handled the legacy plugin=obfs-http case,
so any ss:// link using ws/xhttp/grpc transport silently fell back to TCP and
failed to connect.

Added a call to the shared getItemFormQuery() before the existing obfs check
same function VLESS/VMess/Trojan already use for this. Legacy obfs links still
work since that check runs after and overrides network/headerType as before.

Also fixed toUri(), which passed null for the query dict and dropped these
fields when re-exporting a profile back to a share link.

Tested with an ss:// link using type=xhttp connects now, failed before :)


2dust:
感谢 PR
但是你这个属于 VLESS  的分享规范，不是 ss 的，所以没有订阅是支持这种行为的。

Issue #5920, opened 2026-07-16
Bug: Wrong profile selected in ProxyChain when picking from dropdown


akellavk:
## Description
When adding or editing a ProxyChain configuration, selecting a profile from the dropdown list results in saving a different profile than the one that was clicked. This happens consistently when the dropdown list is filtered or when selecting from a non-first position.

## Steps to Reproduce
1. Open ProxyChain configuration (add new or edit existing)
2. Click on the second member dropdown
3. Select a specific profile from the list (e.g., "🇸🇪 Hysteria")
4. Save the configuration
5. Observe that a different profile is saved (e.g., "ProxyMtpVPN-467")

## Expected Behavior
The profile that was clicked in the dropdown should be saved.

## Actual Behavior
A different profile from the list is saved instead. 

## Environment
- Device: Xiaomi Mi 11 Ultra
- Android Version: 13
- App Version: 2.2.6

2dust:
等我们完成 compose 基础开发再处理

PR #5921, opened 2026-07-16
Fix wrong profile selection in ProxyChain dropdown (#5920)

akellavk:
## Problem
When selecting a profile from the ProxyChain member dropdown, a different profile was saved instead of the one that was clicked. This happened consistently when the user interacted with the filtered dropdown list. Reported in issue #5920 

## Root Cause
The issue was in `ServerProxyChainMemberAdapter.kt`. The code used `suggestions[selectedIndex]` to retrieve the selected profile, but `selectedIndex` from `setOnItemClickListener` refers to the index in the **filtered adapter list**, not the original `suggestions` array.

When `AutoCompleteTextView` filters the dropdown (which happens as users type or interact with it), the indices change, causing the wrong profile to be selected.

## Solution
Changed the implementation to use `parent.getItemAtPosition(selectedIndex)` which retrieves the actual selected value directly from the adapter, regardless of filtering state.

### Before
```kotlin
holder.binding.spMemberRemark.setOnItemClickListener { _, _, selectedIndex, _ ->
    members[currentPos] = suggestions[selectedIndex].trim()
    // ...
}
```

### After
```kotlin
holder.binding.spMemberRemark.setOnItemClickListener { parent, _, selectedIndex, _ ->
    val selectedValue = parent.getItemAtPosition(selectedIndex) as? String ?: return@setOnItemClickListener
    members[currentPos] = selectedValue.trim()
    // ...
}
```

## Testing
Tested by:
1. Creating a new ProxyChain configuration
2. Selecting different profiles from the dropdown for multiple members
3. Verifying that the selected profiles are correctly saved
4. Editing existing ProxyChain configurations and changing member selections

All selections now work correctly and save the intended profile.

## Changes
- Modified `ServerProxyChainMemberAdapter.kt`: Fixed profile selection logic in `setOnItemClickListener`

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/ServerProxyChainMemberAdapter.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/ServerProxyChainMemberAdapter.kt
@@ -31,7 +31,6 @@ class ServerProxyChainMemberAdapter(
     }
 
     override fun onBindViewHolder(holder: MemberViewHolder, position: Int) {
-        val adapterPos = holder.bindingAdapterPosition.takeIf { it != RecyclerView.NO_POSITION } ?: position
         val value = members[position]
         holder.binding.tvMemberIndex.text = (position + 1).toString()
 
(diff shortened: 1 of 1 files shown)


2dust:
已经迁移到 compose

PR #5926, opened 2026-07-17
Random and Round-Robin strategies can exclude dead outbounds

eliotcougar:
This is a small and focused PR aiming to improve usefulness of `PolicyGroups`.

`Random` and `roundRobin` strategies for `PolicyGroups` do not exclude dead outbounds by default. It makes them almost useless in my opinion, because if a `PolicyGroup` contains dead outbounds, you are guaranteed to get intermittent connection errors. It is a bit obscure, but apparently, according to `xray-core` source, these strategies will take into account the health of the outbounds using regular Observatory data if the balancer includes a `fallbackTag`... Any `fallbackTag`.

I don't think it's documented. This is absolutely unintuitive and strange, but it actually works.

This PR adds "Test outbounds" toggle to `random` and `roundRobin` `PolicyGroups`. I made it default on, because it makes more sense in my opinion. It tells config builder to add an Observatory and put the first member of the balancer as the `fallbackTag` of the same balancer. Or the user can set an explicit outbound to be used as the fallback if they want (including `block` and `direct`).

I wasn't sure whether to add fallback field to all types of `PolicyGroups`. I saw a feature request asking for it, I think... Technically, the feature can be simplified to only a single toggle: ON: use only live outbounds; OFF: use all outbounds, disable observatory.


2dust:
代码看上去没有明显问题了。但是实际使用效果如何不确定；你自己使用后没有问题请告知下。

如果你的节点非常多使用策略组，可能会比较耗电和资源，xray 的策略组比较简单和原始。

eliotcougar:
> 代码看上去没有明显问题了。但是实际使用效果如何不确定；你自己使用后没有问题请告知下。
> 
> 如果你的节点非常多使用策略组，可能会比较耗电和资源，xray 的策略组比较简单和原始。

I constantly run policyGroups over a set of approximately 8-10 outbounds. It's a leastPing policyGroup at the moment, after I discovered a major issue with leastLoad strategy in xray-core (https://github.com/eliotcougar/xray-routing-through-unreliable-outbounds)
Constantly running GPS location emulator consumes more power than v2rayNG...

Also, xray-core dev agreed that calling leastLoad strategy "Most Stable" is very misleading as it wasn't their intended meaning. Maybe you can quickly patch that while you work on Composer migration.

V2rayNG/app/src/main/java/com/v2ray/ang/core/CoreConfigManager.kt
```diff
@@ -378,7 +378,6 @@ object CoreConfigManager {
         val fallbackTag = if (strategyType.supportsObservatory && resolvedOutbound.profile.policyGroupTestOutbounds != false) {
             resolvedOutbound.profile.policyGroupFallbackTag
                 ?.takeIf { it.isNotEmpty() && it != AppConfig.TAG_PROXY }
-                ?: membersToAdd.first().tag
```
eliotcougar: If you remove this, what outbound does it set in the xray-core config if the user doesn't set it explicitly in the policyGroup settings? I made that field optional intentionally.

2dust: membersToAdd.first().tag 不应该被添加为 fallbackTag 
因为都要 fallback 了，说明所有的都已经无效了，那么添加 first 也是无意义的，是否正确?

用户应该要为自己添加的 fallbackTag  负责，保证能用

2dust:
> > 代码看上去没有明显问题了。但是实际使用效果如何不确定；你自己使用后没有问题请告知下。
> > 如果你的节点非常多使用策略组，可能会比较耗电和资源，xray 的策略组比较简单和原始。
> 
> I constantly run policyGroups over a set of approximately 8-10 outbounds. It's a leastPing policyGroup at the moment, after I discovered a major issue with leastLoad strategy in xray-core (https://github.com/eliotcougar/xray-routing-through-unreliable-outbounds) Constantly running GPS location emulator consumes more power than v2rayNG...
> 


如果你已经测试过没有问题，那下一版发布后再来合并，因为下一版可能会有大量的bug 

> Also, xray-core dev agreed that calling leastLoad strategy "Most Stable" is very misleading as it wasn't their intended meaning. Maybe you can quickly patch that while you work on Composer migration.

没有明白这个意思？需要做什么

V2rayNG/app/src/main/java/com/v2ray/ang/core/CoreConfigManager.kt
eliotcougar: The fallbackTag has to be populated with something in order to enable health-checking. Basically, it needs some default value. If the user toggles the switch, but doesn't set an explicit outbound, the switch will lie, as xray-core won't check health.
When all outbounds are dead it doesn't really matter which one is in the fallbackTag, unless the user decides to set it to a presumably working outbound outside of members list.

The main reason is the weird roundabout way it is implemented in xray-core. Why couldn't they just add a simple parameter?

eliotcougar:
> 没有明白这个意思？需要做什么

I'm sorry. I see that it is already called `Least Load` in the policy group type.
I confused it with v2rayN - leastLoad strategy is called "Most Stable" there.

2dust:
请解决下冲突，然后合并

eliotcougar:
As expected, it doesn't work when fallbackTag is empty... The final fix commit is coming shortly...

PR #5938, opened 2026-07-22
Two-row status bar displaying current connection

eliotcougar:
This PR depends on a small additional function in https://github.com/2dust/AndroidLibXrayLite/pull/202

The status bar is split in two rows. The first row always displays "Not connected", "Connected", or "Connected via %s". If a leastPing or leastLoad policyGroup is selected, the %s is the actual outbound currently selected by the xray-core balancer. Otherwise, it is the outbound's or the random/roundRobin policyGroup's name.

The second row shows "Tap to check connection" or the test results.

Screenshot_20260722-152249


2dust:
感谢 pr
但是这个 pr 不划算，做的太复杂，只是一个简单的功能

eliotcougar:
> 感谢 pr 但是这个 pr 不划算，做的太复杂，只是一个简单的功能

I guess, most of the complexity comes from me trying to not waste CPU cycles on updating the currently connected outbound when it is not necessary. The core feature is just taking advantage of the proposed API request in the xray lib.

Knowing what outbound is actually used at the moment is useful for diagnostic.

What if I propose a simpler alternative:
To fetch the principal outbound when performing manual connection check and display the outbound that was used together with the measured latency. That simplifies the PR to the core of it - calling `CoreServiceManager.getActiveBalancerTarget(): String` and only if the user is connected through a policyGroup.

2dust:
手机等移动设备，这种功能都是不必要，因为会浪费资源；
不可能做一个像 PC 版本实时能看到很多信息的功能；真要做类似的功能，在这个小空间内也做不好，应该要做一个 dashboard 页面出来

PR #5950, opened 2026-07-25
Parse Xray TLS parameters in Shadowsocks links

vladon:
## Summary

- parse Xray transport and TLS query parameters from modern Shadowsocks URIs
- keep the existing SIP003 `plugin=obfs-local;obfs=http` mapping intact
- add regression coverage for Xray-native TLS, standard SIP002 links, and obfs-local links

## Root cause

`ShadowsocksFmt.parseSip002` decoded the query only for the `plugin` key. It ignored `type`, `security`, `sni`, `alpn`, and `fp`, even though `ProfileItem` already has those fields and `CoreOutboundBuilder.toOutboundShadowsocks` already consumes them when building `streamSettings`.

As a result, a link generated for a Shadowsocks inbound with Xray-native TLS was imported as plain Shadowsocks and failed during the TLS handshake.

This change reuses the same `getItemFormQuery` mapping already used by other Xray-aware URI parsers. Standard SIP002 links without these query parameters remain unchanged. The additional keys are treated as an explicit Xray extension; they are not claimed to be part of SIP002.

## Validation

- TDD regression: the TLS import test failed before the parser change and passed afterward
- `ShadowsocksFmtTest`: 17/17 passed for `fdroidDebug`
- `ShadowsocksFmtTest`: 17/17 passed for `playstoreDebug`
- `assembleFdroidDebug`: passed
- `./gradlew test --continue`: the project compiles and the new tests pass; the command still reports the same two pre-existing `UtilsTest` failures in each flavor (`test_isIpAddress` and `test_IsIpInCidr`) that were present on the untouched upstream baseline

## Related issue

Client-side counterpart to https://github.com/MHSanaei/3x-ui/issues/6094. That issue also tracks 3x-ui's own lossy Shadowsocks importer and raw-export compatibility behavior, which remain separate from this v2rayNG parser fix.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/fmt/ShadowsocksFmt.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/fmt/ShadowsocksFmt.kt
@@ -50,6 +50,7 @@ object ShadowsocksFmt : FmtBase() {
 
         if (!uri.rawQuery.isNullOrEmpty()) {
             val queryParam = getQueryParam(uri)
+            getItemFormQuery(config, queryParam)
             if (queryParam["plugin"]?.contains("obfs=http") == true) {
                 val queryPairs = HashMap<String, String>()
                 for (pair in queryParam["plugin"]?.split(";") ?: listOf()) {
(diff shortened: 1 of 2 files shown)


2dust:
请问 Shadowsocks  在什么时候会用到 tls ？
为什么不用 VLESS ？
且 Shadowsocks  已经不再维护新功能了

PR #5955, opened 2026-07-26
Fix silent drop of configs with illegal URI fragments

ve3xone:
Foreign subscriptions and clipboard pastes often leave raw characters in the remark (the part after `#`) that a URI can't hold: spaces, a second `#`, `|`, cyrillic, emoji. `java.net.URI` throws on those, the per-protocol parsers catch it and return null, and the whole batch of configs disappears with nothing shown to the user.

An example that gets dropped right now:

```
vless://uuid@host:443?type=grpc&security=reality#🇷🇺 Белый интернет #1 | Все операторы
```

Second problem: some panels hand configs out space-separated on one line instead of one per line. `lines()` sees that as a single entry, so everything after the first config is lost.

The fix touches two things in `Utils`:

`fixIllegalUrl` only rewrites the fragment now. Anything that isn't valid in an RFC 3986 fragment gets percent-encoded (non-ascii as utf-8), and existing `%XX` is left as-is so re-running it is a no-op. A url without `#` comes out unchanged, so subscription-url checks aren't affected. Since every uri-based protocol (vmess, vless, trojan, ss, socks, wireguard, hysteria2) goes through this, one place covers all of them.

`splitConfigEntries` replaces `lines()` on the config-import path. It splits on newlines and on whitespace before a known scheme, so both newline- and space-separated blobs end up as the same list. `http(s)://` is skipped on purpose so subscription urls stay in one piece.

Added tests in `UtilsTest` for the remark round-trip (emoji/cyrillic/inner-`#`/`|`), idempotency, a bare `%`, and the split cases. The app's own exports already percent-encode the remark, so they still round-trip fine.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/handler/AngConfigManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/handler/AngConfigManager.kt
@@ -243,8 +243,7 @@ object AngConfigManager {
 
             // Parse all configs first (no I/O during parsing)
             val configs = mutableListOf<ProfileItem>()
-            servers.lines()
-                .distinct()
+            Utils.splitConfigEntries(servers)
                 .reversed()
                 .forEach {
                     val config = parseConfig(it, subid, subItem)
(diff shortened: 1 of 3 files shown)


2dust:
感谢贡献！解决配置导入时静默丢弃的问题确实很有意义，但 
Utils 职责不清：Utils 作为通用工具类，不应该写死协议列表（CONFIG_SCHEMES）等业务逻辑，这破坏了模块的独立性。

容错边界问题：源头格式非法时，在解析端盲目进行妥协和容错，可能会引发更多不可预期的边界 Bug，导致代码越来越复杂。

就是我们应该解决源头，而不是让终端来妥协

PR #5971, opened 2026-07-30
Validate subscription update interval input

eliotcougar:
## Summary

Parse the subscription update interval safely and show the existing validation error when the value is empty, malformed, or outside the `Long` range.

This prevents saving the subscription from crashing with `NumberFormatException`.


2dust:
https://github.com/2dust/v2rayNG/commit/8421d2300b5d213d361a5f61e979f0f05335abe2

PR #5975, opened 2026-07-30
Use typed server menu actions and explicit delete dialogs

eliotcougar:
## Summary

- Replace the profile Share/More menus' positional string arrays and index arithmetic with typed `ServerMenuAction` entries.
- Route profile deletion from every menu through the same confirmation-preference path as the row Delete action.
- Replace the generic confirmation helper with an explicit `DeleteConfirmDialog`.
- Give destructive confirmations a clear Delete label, trash icon, and safe initial focus on Cancel.
- Describe the actual deletion target with complete localized sentences.
- Label the QR dialog's action as Close instead of OK.

## Why

The previous profile menus kept their visible labels and behavior in separate structures. Filtering complex profiles relied on `takeLast()`, `skip`, and adjusted integer positions before dispatching an action. This made the behavior sensitive to menu ordering and allowed the More-menu Delete action to bypass the confirmation path used elsewhere.

Destructive dialogs also used a generic confirmation API and a generic “Confirm delete?” message. The dialog did not communicate what would be removed, and an ambiguous positive action such as OK gives the user less information at the point where a mistake is irreversible.

### Type-safe menu actions

Each displayed menu entry now carries its own action and localized label resource. Filtering changes the list of typed actions rather than changing the meaning of integer positions. This removes the parallel label/action contract and makes future menu edits substantially safer to review.

### One deletion path

Deleting a profile from its row or from the More menu now observes the same confirmation preference. The menu can no longer accidentally dispatch deletion directly while another entry point asks for confirmation.

### Explicit destructive semantics

`DeleteConfirmDialog` encodes the behavior expected from a destructive confirmation:

- the affirmative action is explicitly labeled Delete;
- a trash icon reinforces the destructive result without replacing the text label;
- Cancel receives initial keyboard/D-pad focus;
- callers only provide the target-specific message and callbacks.

Keeping this API specific prevents unrelated confirmation dialogs from inheriting destructive styling or defaults and makes destructive call sites easy to identify during review.

### Clear, localization-safe messages

Confirmations now distinguish profiles, policy groups, routing rules, subscription groups, proxy-chain members, asset files, and asset sources. Bulk profile actions also state that they affect the profiles currently shown, and asset deletion includes the asset name.

These are complete sentences in every existing locale rather than a generic format string with an inserted noun. That allows translators to choose the correct word order, grammatical case, and gender for each target.

Populated proxy-chain member rows require confirmation before removal. Empty member rows remain immediately removable because they contain no configured member.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/compose/Dialog.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/compose/Dialog.kt
@@ -39,29 +39,69 @@ import androidx.compose.ui.unit.dp
 import com.v2ray.ang.R
 
 @Composable
+fun ConfirmDialog(
+    title: String? = null,
+    message: String,
+    confirmText: String = stringResource(android.R.string.ok),
+    dismissText: String? = stringResource(android.R.string.cancel),
+    confirmIcon: @Composable (() -> Unit)? = null,
+    onConfirm: () -> Unit,
+    onDismiss: () -> Unit
+) {
+    val confirmFocusRequester = remember { FocusRequester() }
+    val dismissFocusRequester = remember { FocusRequester() }
+
+    LaunchedEffect(dismissText) {
+        if (dismissText != null) dismissFocusRequester.requestFocus()
+        else confirmFocusRequester.requestFocus()
+    }
+
+    AlertDialog(
+        onDismissRequest = onDismiss,
+        title = title?.let { { Text(it) } },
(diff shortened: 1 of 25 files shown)


eliotcougar:
Maybe we can even remove the "Delete config confirmation" from settings entirely and make all destructive actions require confirmation by default?

2dust:
@copilot resolve the merge conflicts in this pull request

PR #5977, opened 2026-07-30
Fix tunnel hang on LTE -> WiFi handover

vemneyy:
When switching from mobile data to WiFi tunnel would stop working and stay dead until the service was restarted by hand. The other direction WiFi to LTE recovered on its own no intervention needed

Here is trimmed logcat excerpt captured on the current mainline release right during a mobile-to-WiFi switch:
```
07-30 23:36:05.828  initializing core...
07-30 23:36:06.248  Starting core successfully
07-30 23:36:06.249  [Warning] core: Xray 26.6.27 started
07-30 23:36:06 – 23:36:18   from tcp:… accepted tcp:… [tun >> proxy]
                            (normal traffic, DNS-over-HTTPS to 1.1.1.1 going through fine)
07-30 23:36:18 – 23:36:41   [Error] app/dns: failed to retrieve response for dns.wechat.com. >
                            Post "https://1.1.1.1/dns-query": context deadline exceeded
                            (repeats dozens of times, every dns-query attempt times out)
07-30 23:36:27.451  StartCore-Manager: Failed to measure delay
                     go.Universe$proxyerror: Get "https://redirector.googlevideo.com/report_mapping...":
                     context deadline exceeded
07-30 23:36:39.453  StartCore-Manager: Failed to measure delay
                     Get "https://www.google.com/generate_204": context deadline exceeded
07-30 23:36:41.631  StartCore-Manager: Failed to measure delay
                     Get "https://redirector.googlevideo.com/report_mapping...": context deadline exceeded
```

The key fact here is the tunnel itself came up fine (core started, traffic flowed normally for the first ~12 seconds) and then just stopped resolving or answering anything DNS timing out the manual connection test timing out too for the rest of the capture with nothing recovering on its own. No crash, no error that would restart anything, it just sits there dead until you kill the service yourself

Androids LTE -> WiFi handover is make-before-break. System brings WiFi up while LTE is still alive and only tears LTE down after a linger timer, once its confident the new network is stable

The connection to the VPN server established over LTE doesnt get an explicit reset because of this. It simply stops carrying data while from the cores point of view its still "established." Nothing tells the core the network changed so it keeps holding on to a dead connection

The reverse handover (WiFi -> LTE) is break-before-make. Old network disappears immediately so sockets fail with an error and the core redials on its own

The app already had a NetworkCallback registered but only to call setUnderlyingNetworks() for the systems bookkeeping (metering, VPN transport reporting). It received the handover notification correctly it just never told the core to act on it. The fix hooks a core rebuild into that same, already-existing callback

Mine fix tracks which network is actually carrying the tunnel. When that network changes (a real handover) core rebuild is scheduled on separate task with a short delay (~1 second) so it doesnt react to brief blips

Built and tested personally on a real device LTE <-> WiFi switching works correctly in both directions with no manual service restart

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/contracts/ServiceControl.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/contracts/ServiceControl.kt
@@ -1,6 +1,7 @@
 package com.v2ray.ang.contracts
 
 import android.app.Service
+import android.net.Network
 
 interface ServiceControl {
     /**
@@ -25,4 +26,13 @@ interface ServiceControl {
      * @return True if the socket is protected, false otherwise.
      */
     fun vpnProtect(socket: Int): Boolean
+
+    /**
+     * Declares the networks the tunnel runs on top of.
+     * Only meaningful for the VPN service, the other run modes have no interface to report.
+     *
+     * @param networks The upstream networks, null to let the system pick.
+     * @return True if the networks were accepted.
+     */
+    fun setUnderlyingNetworks(networks: Array<Network>?): Boolean = false
 }
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/core/CoreServiceManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/core/CoreServiceManager.kt
@@ -31,6 +31,7 @@ import com.v2ray.ang.util.Utils
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
+import kotlin.jvm.Volatile
 import libv2ray.CoreCallbackHandler
 import libv2ray.CoreController
 import libv2ray.ProcessFinder
@@ -45,6 +46,9 @@ object CoreServiceManager {
     private var processFinder: XrayProcessFinder? = null
     private var browserDialer: IDialerService? = null
     private var networkMonitor: NetworkMonitor? = null
+
+    @Volatile
+    private var isReloading = false
 
     /** Tun descriptor the core was started with, null in the proxy only and root run modes. */
     private var currentVpnInterface: ParcelFileDescriptor? = null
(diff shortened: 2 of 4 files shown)


vemneyy:
@2dust please review pr

eliotcougar:
I already have the fix, and I currently run it on my own phone, but my PR #5909  wasn't accepted due to complexity. It does slightly more to keep connection stable.
I didn't have to press "Restart service" for many days.

vemneyy:
> I already have the fix, and I currently run it on my own phone, but my PR #5909 wasn't accepted due to complexity. It does slightly more to keep connection stable. I didn't have to press "Restart service" for many days.

Hi, I looked at your solution but its really too big. The project code already has a network change handler but it doesnt have handover handling. I tried to minimize it as much as possible changes in the code. Therefore, there is no need to write additional handlers, only need to supplement the old handler

2dust:
感谢，
LTE $\rightarrow$ Wi-Fi Handover 导致套接字悬挂的问题确实需要解决。
不过从架构维护角度来看，直接在现有回调里塞逻辑会让 VPN 核心服务过于臃肿。

后续考虑将网络监听逻辑独立解耦，新建专用的网络监控类（如 NetworkMonitor）：

NetworkMonitor：只负责监听网络拓扑变化、判定 Handover 以及防抖（Debounce）处理。
VpnService：只负责订阅高层事件并在需要时重建核心。

这样单一职责的设计能让代码更清晰

vemneyy:
@2dust refactored it into separate NetworkMonitor class as you suggested:

NetworkMonitor: registers the callback, tracks which network carries the tunnel,
decides a handover happened and debounces it
CoreVpnService: subscribes to that and rebuilds the core

eliotcougar:
Excellent!.. Maybe I can move some of the connection stability code I have to the new NetworkMonitor now...

2dust:
网络  的重载逻辑不应该与 CoreVpnService 强耦合。
在 onlyProxy 模式，网络切换同样会导致 Core 内已建立的链接断开/悬挂。

建议：

1. 保持 CoreVpnService 的单一职责：仅处理 VPN 特有 API（如 setUnderlyingNetworks）。
2. 尽量将 NetworkMonitor 的调度或回调逻辑放至 CoreServiceManager 中或其他里面，或者发送 message 让 CoreServiceManager 中的服务接收。
3. CoreServiceManager  中的代码也不应该大改，而是收到 NetworkMonitor 的消息后做重启即可

vemneyy:
@2dust `CoreVpnService.kt` now has all network handling removed and only setUnderlyingNetworks is left. `NetworkMonitor.kt` is unchanged, it detects the handover and calls back, nothing more. As of this commit `CoreServiceManager.kt` owns the monitor and does the reload and `ServiceControl.kt` has a default no-op setUnderlyingNetworks, so CoreProxyOnlyService and CoreRootService are unchanged and covered too that handles onlyProxy

2dust:
感谢，已经达到目标，可以合并了

Issue #5979, opened 2026-07-31
Suggestions to Improve V2rayNG Functionality and Interface


Fuad-Sabaghi:
Hello,

I'm a V2rayNG user, and I'd like to share a few suggestions that I believe could improve both the functionality and the overall user experience of the app.

First, when managing a large group of configurations (for example, around 500 configs), it's very easy to accidentally tap Real Delay, which immediately starts testing all configurations. This can interrupt the workflow, especially when the intention was simply to sort the configs. I suggest placing Real Delay Config and Sort by Test Results as separate icons in the top toolbar, similar to the Windows version. Separating these actions would help prevent accidental taps and make managing large numbers of configurations much more convenient.

My second suggestion is about the app's design. The current UI and color scheme have remained the same for a long time. I would really appreciate it if future updates introduced a refreshed design with a more modern color palette, such as combinations of gray, cherry, and white.

Finally, V2rayNG is undoubtedly one of the best applications of its kind. However, it would be even better if it had a dedicated Home page similar to V2Box. This page could display useful information such as the connection duration, uploaded and downloaded data, and the location of the connected server. I believe this would greatly improve the overall user experience.

Thank you very much for your time and for your hard work on this excellent application.

2dust:
请使用最新版，测试菜单有变动
配色如果有人能 pr 我们可以接受
Home page 不是我们想要的功能，目前没有计划；因为目前大部分主页都是放一些无意义的内容，浪费资源

PR #5984, opened 2026-07-31
Allow cancelling batch tests from the status bar

eliotcougar:
## Summary

- make the bottom status area cancel an active batch latency test instead of starting a current-server connectivity check
- show live test-task progress together with a localized “Tap to stop” hint
- ignore queued progress updates after cancellation so the normal connection status remains visible
- add the new status text to every existing locale

## Why

Issue #5979 describes accidentally starting **Real delay config** for very large profile groups, with no obvious way to stop the lengthy operation.

The test service and view model already supported cancellation, but the bottom status area always dispatched `TestCurrentServer`. This change uses the existing `isTesting` state to route that same status-area tap to `CancelTesting` only while a batch test is active. Normal status taps keep their current connectivity-check behavior.

This addresses the cancellable batch-test part of #5979 without changing how latency tests are scheduled or executed.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainBottomBar.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainBottomBar.kt
@@ -35,6 +35,7 @@ import com.v2ray.ang.ui.compose.colorFabInactiveLight
 fun MainBottomBar(
     displayText: String,
     isRunning: Boolean,
+    isTesting: Boolean,
     isDarkTheme: Boolean,
     onAction: (MainAction) -> Unit
 ) {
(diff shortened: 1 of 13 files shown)


2dust:
现在测试是会显示一个通知信息，可以考虑在通知信息中添加取消，我认为是一个更好的方式。

eliotcougar:
> 现在测试是会显示一个通知信息，可以考虑在通知信息中添加取消，我认为是一个更好的方式。

I may be confused about terminology.
The PR adds "Tap to cancel" right where the in-progress results are being displayed. I called it a "status bar", the area at the bottom. Is that the "notification" you're talking about?

2dust:
系统通知， notification

eliotcougar:
> 系统通知， notification

I've never seen this "test in progress" system notification. Am I missing something?

UPD: I have finally found it... I didn't have notifications enabled...
Yes, I can put the Cancel button there, but I worry about discoverability if the notification is unnoticed. Can we increase the importance of the CORE_TEST notification channel so that it appears more reliably?

PR #5989, opened 2026-08-01
Remove unused legacy server editor

eliotcougar:
Remove the unused legacy `ServerActivity` implementation.

`ServerActivity` is not declared in the application manifest and has no source, navigation, or inheritance references. Profile creation and editing are handled by the active protocol-specific editor activities, so this 524-line implementation cannot be reached at runtime.


2dust:
特意留下来作为代码参考的，等两个版本再删除

eliotcougar:
Oh, alright. I thought that might be why it's still there.

PR #5998, opened 2026-08-01
Fix/network handover stability

eliotcougar:
In addition to #5994 it is a good idea to make the handover idempotent and not reset state on duplicate underlay change events when Android is configured to keep both wifi and cellular connected (aka Adaptive Connectivity).


2dust:
现有的功能经过测试已经很不错了， 此 PR 是不是让事情变得过于复杂了。 
你说的重复的底层事件是指什么？ 比如 wifi 重复的断开和连接？

PR #6000, opened 2026-08-01
Optimize app picker filtering and bulk selection

eliotcougar:
## Summary

- keep the new upstream Coil-based app icon and list-loading path unchanged
- filter app names and package names without allocating uppercase copies for every row
- apply search changes directly and preserve the active query if initial loading finishes later
- compute bulk, automatic, and imported selections once before publishing and persisting them
- prevent selection mutations while another picker operation is still running

## Rationale

Upstream now loads application metadata without icons and retrieves icons asynchronously through Coil. That supersedes the icon and progressive-loading implementation previously carried by this PR, so those changes have been removed.

The remaining work addresses a separate cost in selection handling. The Per-app proxy picker previously implemented invert, automatic selection, and imported selection by repeatedly calling the same single-item add/remove functions. Each changed item published a new set, wrote it to MMKV, and requested a service restart. The amount of state and persistence work therefore grew with the number of matching applications.

This PR builds the final set first, publishes and persists it once, and requests at most one service restart for each user action. Automatic and imported selection calculations run away from the main thread and use the complete installed-app list rather than whichever rows happen to match the current search.

Search also no longer creates uppercase copies of every label and package name on each query change, and the Compose screens call the filter directly instead of launching a new effect for every keystroke.

## Compatibility

- App ordering remains selected apps first, then user apps before system apps, with locale-aware label sorting.
- Bypass mode and forced Google-app handling are unchanged.
- The existing partial package-name matching behavior is intentionally preserved.
- Upstream's `AppInfo`, `AppManagerUtil`, `AppIconFetcher`, Coil dependency, and asynchronous icon rendering are unchanged by this PR.

## Validation

- Focused `AppPickerViewModelTest` passed.
- `:app:compilePlaystoreDebugKotlin` passed.
- `:app:assemblePlaystoreDebug` passed.
- The full Play Store debug unit suite completed 27 tests: 25 passed, with only the two unchanged baseline failures in `UtilsTest.test_isIpAddress` and `UtilsTest.test_IsIpInCidr`.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/compose/Components.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/compose/Components.kt
@@ -160,14 +160,13 @@ fun AppListItem(
     icon: Any?,
     checked: Boolean,
     onCheckedChange: (Boolean) -> Unit,
-    modifier: Modifier = Modifier,
-    enabled: Boolean = true
+    modifier: Modifier = Modifier
 ) {
     val context = LocalContext.current
     Row(
         modifier = modifier
             .fillMaxWidth()
-            .clickable(enabled = enabled) { onCheckedChange(!checked) }
+            .clickable { onCheckedChange(!checked) }
             .padding(horizontal = 16.dp, vertical = 12.dp),
         verticalAlignment = Alignment.CenterVertically
     ) {
@@ -211,7 +210,6 @@ fun AppListItem(
         Checkbox(
             checked = checked,
             onCheckedChange = onCheckedChange,
-            enabled = enabled,
             colors = CheckboxDefaults.colors(checkedColor = colorFabActive)
         )
     }
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/compose/Components.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/compose/Components.kt
@@ -160,13 +160,14 @@ fun AppListItem(
     icon: Any?,
     checked: Boolean,
     onCheckedChange: (Boolean) -> Unit,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true
 ) {
     val context = LocalContext.current
     Row(
         modifier = modifier
             .fillMaxWidth()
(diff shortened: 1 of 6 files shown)


V2rayNG/app/src/main/java/com/v2ray/ang/ui/compose/Components.kt
```diff
@@ -158,53 +162,20 @@ private fun SearchInputField(
 fun AppListItem(
```
2dust, line 157: 是不是应该把 app 相关 compose 独立一个文件？

2dust:
经过实际测试，初次加载确实快了一些，但是快速向下或向上大幅度滑动时，卡顿比较明显，看日志应该是在加载图片。

不说资源占用问题，这个体验感觉还不如优化前

eliotcougar:
> 经过实际测试，初次加载确实快了一些，但是快速向下或向上大幅度滑动时，卡顿比较明显，看日志应该是在加载图片。
> 
> 不说资源占用问题，这个体验感觉还不如优化前

You are correct.
My "slow device testing" only included my TV. It's not possible to "fast scroll" on it.  I didn't catch jankiness when fast-scrolling on Pixel 9 pro because it's fast.

I removed the lazy loading. Now it loads and decodes icons in the background until completion and keeps them loaded. All the other optimizations remain.

V2rayNG/app/src/main/java/com/v2ray/ang/ui/compose/Components.kt
eliotcougar: I am ambivalent on that. Separating them doesn't give much.

V2rayNG/app/src/main/java/com/v2ray/ang/ui/compose/Components.kt
```diff
@@ -160,13 +160,14 @@ fun AppListItem(
     icon: Any?,
     checked: Boolean,
     onCheckedChange: (Boolean) -> Unit,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true
```
2dust: 目前加载已经没有了图片，所以速度很快了。
这里 enabled 可以不要了

PR #6004, opened 2026-08-02
feat: force disable skip certificate verification for subscription nodes

btxbtxbtx:
## Background

Xray-core v26.2.6 removed the `allowInsecure` TLS parameter and migrated it to `pinnedPeerCertSha256` (see [discussion #9460](https://github.com/2dust/v2rayN/discussions/9460)).

Since **June 1, 2026**, any subscription node config that still contains `allowInsecure: true` without certificate pinning causes the Xray core to **fail to start** with:

```
The feature "allowInsecure" has been removed and migrated to "pinnedPeerCertSha256".
Please update your config(s) according to release note and documentation.
```

Keeping `allowInsecure: true` also disables TLS certificate verification entirely, exposing users to MITM attacks.

## Change

This PR adds `forceDisableInsecure()` which is invoked after each subscription update. It:
- Iterates all nodes in the subscription
- Skips `CUSTOM` type nodes (their runtime config comes from raw JSON)
- Sets `insecure = false` on any node currently carrying `insecure: true`
- Writes back only changed nodes (no redundant I/O)

This prevents subscription nodes with the obsolete `allowInsecure` flag from crashing the Xray core on startup.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/handler/AngConfigManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/handler/AngConfigManager.kt
@@ -193,6 +193,10 @@ object AngConfigManager {
             updateConfigViaSubAll()
         }
 
+        if (count > 0) {
+            forceDisableInsecure(subid)
+        }
+
         return count to countSub
     }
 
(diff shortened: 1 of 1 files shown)


2dust:
当前版本保留 allowInsecure 是为了给用户明显的提示。
我们将在最近的时间，会完全把 allowInsecure 的相关代码都移除

2dust:
如果你有时间，也可以 pr 帮我们处理掉 allowInsecure

PR #6020, opened 2026-08-04
ping json servers

shashachkaaa:


eliotcougar:
This looks like it came from a very different branded fork of v2rayNG... You should ask Claude to simplify the branch immensely, remove all unrelated changes, remove unnecessary code, and strip it down to the bare minimum implementation focused on the proposed feature…

PR #6031, opened 2026-08-07
fix hy2 port and fetch not show

anyiln:


2dust:
HY2 不能用 CommonStreamSecurityFields

PR #6025, opened 2026-08-05
fix: localize user-facing and accessibility text (2/3)

eliotcougar:
## Context

Partially fixes #6005.
Fixes #6022.

This is **part 2 of 3** extracted from the original combined localization PR #6021. It follows the merged #6024 (part 1), which aligns the locale catalogs and repairs missing/divergent translations.

The branch has now been rebased onto current `upstream/master` after #6024 merged. Its commit list and Files changed view contain only the five focused part-2 commits.

Focused range: `d8b857665c32f73d34dab220b324f54af336679a...5f7c8dc9c298dc3baf2d6bdabb0f39f5934cda64`

### Localize user-facing runtime text

- Keeps startup and reload diagnostics in English for Logcat while exposing localized UI failures for missing server selections, profile decoding, config-context creation, normal config generation, empty custom configurations, and core startup failures.
- Localizes the empty-detail fallback shown by connection tests.
- Uses a localized `Unknown` fallback for missing country/IP data and an unavailable native-core version; the core-version API is now nullable instead of leaking an English sentinel into the UI.
- Makes permission names (`Camera`, `Notifications`, and `Local network`) resources and formats permission-denied messages so each language can order the permission name naturally.
- Moves malformed-JSON details into a formatted resource while retaining the generic fallback when no detail is available.
- Localizes subscription-update status and progress notifications, including the active subscription/group context.
- Localizes Tasker start/stop descriptions.

### Give TalkBack controls dedicated descriptions

- Adds clearly named `acc_` resources whose wording describes the action or state when spoken aloud.
- Separates accessibility descriptions from visible labels instead of reusing unrelated visual identifiers.
- Covers main navigation and start/stop actions; back, search, add, more, edit, delete, and save actions; subscription and asset actions; routing controls; server/profile/group/custom-config actions; proxy-chain membership; per-app proxy controls; QR and scanner controls; and Logcat refresh/search/copy/share/clear controls.
- Makes the scanner torch description state-aware (`Turn torch on` / `Turn torch off`).
- Keeps visible UI resources intact; these changes apply only to accessibility semantics.
- Removes the misleading `Pull down to refresh` toast shown when Logcat opens now that the screen exposes a dedicated refresh button, and gives that button the spoken description `Refresh log`.

### Complete and review Bakhtiari (BQI) coverage

- Translates every new part-2 user-facing and accessibility resource instead of leaving English placeholders.
[...]

PR #6024, opened 2026-08-05
fix: sync locale resources and translations (1/3)

eliotcougar:
## Context

Partially fixes #6005.

This is part 1 of 3 extracted from the broader localization audit in #6021, following the request to split that PR into independently reviewable changes. This part is limited to synchronizing and correcting the Android resource files, completing their translations, and updating the few code references required by corrected resource identifiers or resource behavior.

### Synchronize every maintained locale

- Align English, Arabic, Bangla, Bakhtiari, Persian, Russian, Vietnamese, Simplified Chinese, and Traditional Chinese to the same 399 named resources.
- Keep every resource at the same physical row, with the same type, order, formatting placeholders, and translatability metadata in all nine files.
- Add resources that were missing from individual locales, including browser-dialer settings, insecure-protocol warnings, unknown-app text, subscription-worker text, automatic subscription-test settings, mode/root settings, request headers, routing fields, browser-dialer modes, and preset rulesets.
- Use localized wording where a dependable translation is available; retain exact Xray and protocol terminology where translating it would change or obscure the meaning.
- Expand localized string arrays to one item per line while preserving cross-locale row alignment.

### Correct identifiers and references

- Correct `title_pref_mux_concurency` to `title_pref_mux_concurrency`.
- Correct `title_pref_mux_xudp_concurency` to `title_pref_mux_xudp_concurrency`.
- Correct `connection_runing_task_left` to `connection_running_task_left`.
- Correct `toast_malformed_josn` to `toast_malformed_json` and update its call site.
- Remove the unused pre-Compose `toast_fragment_not_available` resource.
- Restore the routing outbound-tag placeholder in Compose and substitute each locale's own translation of `remarks` while preserving the literal `proxy`, `direct`, and `block` tags.

### Improve translation accuracy

- Make routing terminology consistent with the corresponding Settings labels, including `routeOnly` behavior.
- Correct awkward or overly literal wording in both English and translated resources.
- Keep Xray's `reject`, `allow`, `skip`, `leastLoad`, `mldsa65Verify`, `echConfigList`, and other core terms semantically exact.
- Correct translations whose previous wording described a different action or setting, including XUDP-over-QUIC handling, current-load selection, certificate verification, destination-domain detection, speed-notification behavior, and remarks regex filtering.
- Fill the remaining Bakhtiari English fallbacks using established vocabulary while leaving strictly technical tokens intact.
- Standardize `URL` and `ALPN` capitalization.

### Exclude fixed technical values from translation

[...]

PR #6044, opened 2026-08-08
fix: honor the selected app locale everywhere (All-in-one)

### LINKED ISSUE #6005: Sync translation file from values

eliotcougar:
## Context and series

Fixes #6005. For real this time.

This is **part 3 of 3** extracted from the original combined localization audit in #6021.

- **Part 1 — #6024 (merged):** synchronized the locale catalogs, filled missing translations, corrected divergent meanings, and marked invariant technical values non-translatable.
- **Part 2 — #6025 (merged):** extracted hardcoded user-facing text, introduced dedicated accessibility descriptions, and localized the resulting resources.
- **Part 3 — this PR:** makes the selected app language authoritative in every Android context, keeps locale-neutral data across service/UI boundaries, and contains the final translation and wording polish.

## Root cause: why present translations looked missing

The affected resources were present and correctly named. The bug came from **which `Context` resolved them**.

The previous implementation stored the chosen language in MMKV and wrapped `BaseComponentActivity` with a custom locale configuration. That changed resources obtained from the activity, including most direct Compose `stringResource(...)` calls, but it did not establish an application-wide per-app locale.

Several kinds of text were produced elsewhere:

- `AndroidViewModel` and repository code used the raw `Application` context.
- app-picker background work switched to `applicationContext`.
- VPN, proxy-only, root, test, subscription-update, and Quick Settings services had their own contexts.
- connection-test results and other status values were formatted in the service/daemon layer and sent to the UI as finished strings.
- some localized strings were cached in retained UI state before an activity recreation.
- dialog actions used `android.R.string.ok` and `android.R.string.cancel`, which belong to framework resources rather than the app's locale catalogs.

Those contexts continued to use the **device locale**. A single Compose screen could therefore combine correctly localized strings from its activity with Persian, English, or another device-language string created outside that activity. This explains both otherwise puzzling observations from #6021:

1. only some strings were wrong on a screen; and
2. changing the device language changed those strings even though the app language and the locale XML entry did not change.

The `values-bqi-rIR` resource directory was not itself the cause. `bqi-rIR` is the Android resource-qualifier form; the runtime per-app locale API instead receives the BCP 47 language tag `bqi-IR`. The old code treated the stored resource-style code as the entire locale mechanism and never synchronized it with Android's app-locale APIs.

This implementation follows Android's current [per-app language guidance](https://developer.android.com/guide/topics/resources/app-languages) and uses [`ContextCompat.getContextForLanguage`](https://developer.android.com/reference/androidx/core/content/ContextCompat#getContextForLanguage(android.content.Context)) for resource access outside an `AppCompatActivity`.

### Use the platform and AndroidX app-locale APIs

- Changes the Compose activity base from `ComponentActivity` to `AppCompatActivity`, as required for backward-compatible `AppCompatDelegate.setApplicationLocales(...)` handling.
- Adds a centralized `AppLocaleManager` for setting, reading, migrating, and applying the selected locale.
- Separates the existing stored setting code from the proper runtime language tag:
  - `zh-rCN` -> `zh-CN`
  - `zh-rTW` -> `zh-TW`
  - `bqi-rIR` -> `bqi-IR`
- Preserves the existing MMKV preference so upgrades do not discard the user's chosen language.
- Synchronizes changes made through either the in-app picker or Android 13+ system app-language settings.
- Performs the documented one-time handoff from the legacy preference to AppCompat/framework locale storage.

### Declare supported languages correctly

- Enables generated `LocaleConfig` support and explicitly filters it to the nine maintained locales.
- Declares English as the unqualified resource locale through `resources.properties`.
- Adds AppCompat locale auto-storage for Android 12 and older.
- Makes the language list available in Android 13+ system settings without maintaining a second manual locale-config file.
- Uses an AppCompat parent theme required by the new activity base while preserving the existing day/night window styling.

### Apply the locale outside activities

- Resolves `Application` and `BaseViewModel` resources through a locale-aware context.
- Applies the selected locale to VPN, proxy-only, root, test, subscription-update, and Quick Settings service contexts.
- Uses localized resources in `MainRepository`, including the synthetic **All** group.
- Captures and refreshes the localized unknown-app label around app-picker background work.
- Removes the former custom `MyContextWrapper`; locale behavior now has one AndroidX/platform-backed owner.

## Keep core and service data locale-neutral

Part 2 initially exposed more localized failures, but subsequent upstream architecture deliberately removed Android resources from `/core`. This PR preserves that separation.

- `CoreServiceManager` no longer formats connection-test UI sentences.
- The service sends a locale-neutral `ConnectionTestResult` containing delay, diagnostic detail, country code, and IP address.
- `MainViewModel` formats that result only at the UI boundary using the currently selected app locale.
- The raw technical error remains suitable for Logcat while the visible sentence and empty-detail fallback are localized.
- Country codes remain untouched; only the missing-value label is localized.
- Main-screen status is represented as semantic `MainStatus` state instead of retained localized strings, so activity recreation or a locale change cannot leave stale text behind.
- update-check and Logcat-sharing exceptions are logged with their diagnostic detail, while users see a stable localized failure message.

### Units and connection status

- Localizes the millisecond unit used by TCPing and real-delay results in profile rows.
- Formats the connection-test result with that same localized unit.
- Rephrases the success result naturally in every locale.
- Clarifies the connected-state prompt and allows a line break in locales where the combined text is too long.
- Uses a localized `Unknown` value when the native-core version, country, or IP value is unavailable instead of leaking an English sentinel.

### Framework and cached UI text

- Replaces framework `OK` and `Cancel` resources with app-owned translations in dialogs and notifications.
- Recomputes locale-sensitive app-picker and main-screen state instead of retaining strings from the old language.
- Keeps core and service messages structured until the presentation layer can apply the active locale.

### Settings and menu wording

- Clarifies that per-app proxy auto-selection selects **apps that should be proxied**.
- Shortens **Subscription group settings** to **Subscriptions** across locales.
- Localizes “format” in the DNS hosts title while retaining the literal `domain:address,…` syntax.
- Standardizes `URL` capitalization.
- Corrects stale Arabic labels that still described numbered workflow steps.
- Reviews the complete locale catalogs for awkward, overly literal, or calque-like wording while preserving xray-core identifiers, protocol names, raw keys, and other technical terms.

### Routing clarity

- Moves the rule-combination explanation out of individual fields so it appears once.
- States the actual behavior: populated domain, IP, port, network, source, protocol, and process conditions are combined with **AND**; rules are evaluated top-to-bottom and the first matching rule applies.
- Leaves only the comma-separated-entry hint inside the applicable fields.
- Applies the same explanation in all maintained locales.

### Subscription proxy-chain clarity

- Renames the two bottom subscription fields to **Entry proxy** and **Exit proxy**.
- Adds localized supporting text below each field explaining that the selected proxy is inserted before or after every profile in the subscription as the entry or exit link in that profile's proxy chain.
- Keeps the mapping aligned with the actual chain construction: the stored previous profile is the entry hop and the stored next profile is the exit hop.

## Resource consistency

- English, Arabic, Bangla, Bakhtiari, Persian, Russian, Vietnamese, Simplified Chinese, and Traditional Chinese each contain exactly **446 resources**.
- All catalogs retain the same identifiers, resource types, order, physical row alignment, translatability metadata, formatter metadata, and format placeholders.
- Non-translatable values are identical in every catalog.
- String arrays remain unwrapped with matching rows.

## Impact

- Selecting a language in v2rayNG now affects activities, Compose, ViewModels, repositories, services, notifications, the Quick Settings tile, and UI text received from the daemon path.
- Language change applies immediately without the need to restart the app.
- Android 13+ system language settings and the in-app picker remain synchronized.
- Existing users retain their selected language during migration.
- Mixed app-language/device-language screens no longer occur merely because a string was created outside an activity.
- The core remains free of Android resource dependencies.
- The final UI text is shorter, clearer, and more consistent across all nine locales.


hosseinabaspanah:
@eliotcougar Great and thank you for your help, and for addressing my report and improving the program and the program languages.

2dust:
感谢
但是此 pr 又是存在一个很难 review 的问题，ai 处理了太多东西。
还是要拆分成多份 pr

1. AppLocaleManager
2. string 调整
3. 其他

V2rayNG/app/src/main/java/com/v2ray/ang/core/CoreNativeManager.kt
```diff
@@ -56,14 +56,14 @@ object CoreNativeManager {
     /**
      * Get V2Ray core version.
      *
-     * @return Version string of the V2Ray core
+     * @return Version string of the V2Ray core, or null if it cannot be read.
      */
-    fun getLibVersion(): String {
+    fun getLibVersion(): String? {
```
2dust, line 61: 这里不要反复的改了，上次已经回滚过一次

eliotcougar: Version string rendering moved from core to UI (About and Update screens)

eliotcougar:
> 感谢 但是此 pr 又是存在一个很难 review 的问题，ai 处理了太多东西。 还是要拆分成多份 pr

Sigh... It's not AI, it's me...
I though if the majority of changes are about AppLocaleManager, and the rest are cleanup localized in strings.xml, it may be okay.

Alright, I'll split the cleanup work from the #6005 fix.

2dust:
如果不是ai ，也不要在一个 pr 中 提交如此多的修改。
非常难审查和测试

PR #6057, opened 2026-08-10
Mark unsupported insecure profiles as deprecated

eliotcougar:
## Summary

Recent Xray-core changes reject several insecure outbound configurations instead of attempting to connect them. v2rayNG currently leaves those profiles looking usable on the main screen, so users often discover the incompatibility only after connection or real-delay tests fail.

This PR identifies profiles rejected by the Xray-core revision pinned by AndroidLibXrayLite, gives them a light-red background, and replaces their protocol label with `DEPRECATED`. It does not disable editing, sharing, or removal, and it does not change connection behavior.

Related to #6046.

## Compatibility rules

The detector mirrors the insecure-outbound checks from pinned Xray-core commit `5ca6f4b7d4dc`:

- unencrypted VLESS outbounds using no protocol encryption (`none` or empty) to public endpoints;
- Trojan outbounds without TLS or REALITY to public endpoints; and
- generated TLS outbounds that still request the removed `allowInsecure` behavior for VMess, VLESS, Shadowsocks, Trojan, or Hysteria2, unless a `pinnedCA256` certificate fingerprint prevents that setting from being emitted.

Xray's private-endpoint exceptions are mirrored as well. Private and reserved IP ranges, dotless hostnames, and Xray's recognized local/private domain suffixes are not marked merely for lacking transport security.

## Cache and refresh behavior

Compatibility is evaluated once when a profile snapshot is decoded into `ServersCache`; Compose list recompositions reuse the cached Boolean rather than re-parsing addresses and CIDR ranges.

The cache is naturally rebuilt after foreground imports, edits, and list reloads. For scheduled subscription updates, `SubscriptionUpdateService` now emits a targeted data-change event only after a successful profile replacement or a non-empty test run. The main view invalidates the changed subscription and the aggregate **All** cache, immediately rebuilds only the affected group currently on screen, and leaves other large groups to reload lazily when selected. Cache invalidation is synchronized with background preloading so an in-flight preload cannot restore a stale snapshot.

## Validation

- Added 11 focused unit tests covering public and private IPv4/IPv6 endpoints, private domain rules, VLESS protocol encryption, TLS/REALITY, certificate pinning, removed `allowInsecure`, and unrelated protocols.
- `:app:testPlaystoreDebugUnitTest --tests com.v2ray.ang.core.XrayOutboundCompatibilityTest`
- `:app:assemblePlaystoreDebug -PABI_FILTERS=x86_64`

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
@@ -188,6 +188,7 @@ object AppConfig {
 
     const val MSG_SUB_UPDATE_START = 8
     const val MSG_SUB_UPDATE_CANCEL = 81
+    const val MSG_SUB_UPDATE_DATA_CHANGED = 82
 
     /** Notification channel IDs and names. */
     // Use a new ID because Android does not let an app raise an existing channel's importance.
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/core/XrayOutboundCompatibility.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/core/XrayOutboundCompatibility.kt
@@ -1,0 +1,149 @@
+package com.v2ray.ang.core
+
+import com.v2ray.ang.AppConfig
+import com.v2ray.ang.dto.entities.ProfileItem
+import com.v2ray.ang.enums.EConfigType
+import java.net.InetAddress
+
+/**
+ * Mirrors the insecure-outbound validation in the Xray-core revision pinned by
+ * AndroidLibXrayLite (5ca6f4b7d4dc).
+ */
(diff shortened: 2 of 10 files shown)


eliotcougar:
@2dust, feel free to reject it. It is a temporary measure kind of feature until people migrate to more secure service providers.

2dust:
感谢 PR
我们不支持这种做法，没有必要替用户做决定，不要做家长；用户想用什么是自由的，自己为自己负责。

eliotcougar:
> 感谢 PR 我们不支持这种做法，没有必要替用户做决定，不要做家长；用户想用什么是自由的，自己为自己负责。

I agree. However, I still think that deprecation of the insecure configurations by xray-core is not communicated well. The error message given by the app is not specific enough.

PR #6061, opened 2026-08-11
feat: show subscription traffic and expiration

### LINKED ISSUE #6045: subscription expiration time graphically
The app is very good, but it could be improved with some ideas. For example, like the happ app, it could show the subscription expiration time graphically.

Automatic subscription updates also don't work.

Image

tianrking:
Closes #6045

Expose subscription usage and expiration metadata from the standard `subscription-userinfo` response header. The update path now keeps response headers through redirects, persists upload/download/total/expiration values with the subscription, and renders a traffic progress bar plus expiration timestamp in the subscription list when the provider supplies the metadata. Providers that do not send the header retain the existing behavior and hide the optional fields.

Validation:
- `git diff --check`
- Added focused parser coverage for valid, malformed, negative, and unknown header fields.
- The targeted Gradle unit-test command was attempted locally but could not resolve the repository's current Android Gradle Plugin from the offline cache; no build result is claimed.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/dto/SubscriptionUserInfo.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/dto/SubscriptionUserInfo.kt
@@ -1,0 +1,14 @@
+package com.v2ray.ang.dto
+
+data class SubscriptionUserInfo(
+    val upload: Long = 0,
+    val download: Long = 0,
+    val total: Long = 0,
+    val expire: Long = -1
+) {
+    val hasTraffic: Boolean
+        get() = total > 0
+
+    val hasExpiration: Boolean
+        get() = expire > 0
+}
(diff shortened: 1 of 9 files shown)


eliotcougar:
Is this something only one specific provider does, or is it some sort of standard all of them follow?

tianrking:
It is the widely used `subscription-userinfo` response-header format used by many Clash/Xray-compatible subscription services (`upload`, `download`, `total`, and `expire`), but it is not guaranteed by every provider. The implementation keeps the existing behavior and hides the optional UI when a provider does not send that header.

2dust:
感谢你的pr
但是我们已经明确过，不会显示这些内容。
因为 app 本身自己没有统计功能，加了这些数据后也不会在使用时减少流量。
且本应用是针对单个节点的，想做很好的流量统计也是做不到的，既然做不到那干脆就不做

PR #6073, opened 2026-08-14
Add an explicit stop-and-exit action

eliotcougar:
## Summary

- Add a localized Exit item at the bottom of the main More menu on phones and TVs.
- Request service shutdown before removing the application task.
- Have VPN, proxy-only, and root services report when teardown has completed.
- Wait for completion with a five-second fallback timeout.
- Send the stop request even when the UI currently believes the service is stopped, protecting against stale UI state.

## Why

The future Android TV interface needs an obvious way to leave the application using only a remote control.

The action is also useful on smartphones as a convenient way to stop v2rayNG's core service and close its task without navigating to Android's App Info screen.

This is not equivalent to Android's privileged Force stop operation: it does not place the package in the force-stopped state or terminate arbitrary processes. It performs an orderly service shutdown and removes the application task.

## Testing

- `:app:assemblePlaystoreDebug -PABI_FILTERS=x86`

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
@@ -178,6 +178,7 @@ object AppConfig {
     const val MSG_STATE_START_FAILURE = 32
     const val MSG_STATE_STOP = 4
     const val MSG_STATE_STOP_SUCCESS = 41
+    const val MSG_STATE_STOP_COMPLETE = 42
     const val MSG_STATE_RESTART = 5
     const val MSG_MEASURE_DELAY = 6
     const val MSG_MEASURE_DELAY_RESULT = 61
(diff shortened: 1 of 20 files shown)


2dust:
我认为此功能是没有用的。
app 不需要桌面版本的退出功能，切换到后台自动就会被系统处理了。

eliotcougar:
> 我认为此功能是没有用的。 app 不需要桌面版本的退出功能，切换到后台自动就会被系统处理了。

This is an extracted PR from the TV UI branch. Initially, before the language settings were changed so that they applied immediately, I had to force-close and restart the app for the language to apply. On a phone it was easy. On TV the command to force-close the app is hidden far and requires more than a dozen actions to get to it. There may be other situations where the user may want to completely restart the app. When the app switches to background it does not close right away.

This is just an easier way for the user to do a complete app restart reliably without going into app settings and doing force-close. (some people may not even know about force-close)

PR #6076, opened 2026-08-14
Hide Restart service while stopped

eliotcougar:
Much simpler alternative to #6072 

## Summary

- hide **Restart service** from the main More menu while the service is stopped
- keep the existing Start control as the only way to start a stopped service

## Why

Restart is meaningful only for an active service. Making it unavailable while stopped keeps the action semantics explicit and avoids introducing cross-process "restart or start" acknowledgement and fallback logic for a state already represented by the UI.

This also prevents Restart from silently becoming a second Start action.


2dust:
服务重启的作用不止是重启，如果没有启动的时候也是启动服务，所以不应该接受这个 pr

PR #6086, opened 2026-08-15
Safely clean up orphaned server profiles

eliotcougar:
## Summary

- commit replacement profiles and their `SUB_SERVERS_*` group index together under the MMKV multi-process lock
- collect only profile payloads that are provably unreachable from every readable raw group index
- run cleanup once in the background for existing installations and after successful replacement imports or subscription updates
- remove matching full-profile, raw-config, and affinity entries in bulk

## Context

Follow-up to #6085. Related to #6084.

#6084 contained more than 21,000 entries in `PROFILE_FULL_CONFIG`, while only about 4,900 profiles remained referenced by current subscription group indexes. This PR addresses that separate persistence and cleanup problem.

It does not assume that orphan-profile growth caused the reported loss of `SUB_IDS` and `SUB`, and it does not attempt to reconstruct missing subscription metadata. The exact corruption point has not been reproduced, so that remains a separate investigation.

## Cleanup safety

The collector deliberately derives reachability from the raw `SUB_SERVERS_*` indexes instead of `SUB_IDS` or `SUB`, because #6084 shows that those metadata stores may be missing while the group indexes and profiles remain intact.

Cleanup fails closed:

- the selected profile is always preserved
- undecodable profiles are preserved
- profiles whose group index is missing are preserved
- profiles referenced by any group are preserved, even if their embedded group ID disagrees
- the entire pass is skipped if no raw group indexes survive or any group index is unreadable

Replacement ordering is also changed so new payloads and their group index are committed before the previous payloads become cleanup candidates. Selection reassignment completes before collection runs, avoiding an empty or broken intermediate group if a refresh is interrupted.

The one-time migration marker and profile/index updates use the same cross-process lock. The initial scan runs on `Dispatchers.IO`, and cleanup failures are logged without turning an otherwise successful import or subscription refresh into a failure.

## User impact

Existing installations can discard historical profile payloads that are no longer reachable, reducing persistent MMKV and backup growth while preserving ambiguous or potentially recoverable data. Future replacement updates clean their old payloads after the new group is committed.

There are no UI changes, and healthy active profiles keep their current behavior.

## Validation

- `:app:testPlaystoreDebugUnitTest --tests com.v2ray.ang.handler.OrphanProfileCleanerTest` — 7 tests passed
- `git diff --check`

The unit tests cover missing subscription metadata, missing and unreadable group indexes, selected and undecodable profiles, default-group normalization, and cross-group references.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/AngApplication.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/AngApplication.kt
@@ -10,6 +10,10 @@ import com.v2ray.ang.AppConfig.ANG_PACKAGE
 import com.v2ray.ang.handler.AppLocaleManager
 import com.v2ray.ang.handler.SettingsManager
 import com.v2ray.ang.ui.compose.ThemeManager
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.launch
 
 class AngApplication : Application() {
     companion object {
@@ -29,6 +33,8 @@ class AngApplication : Application() {
         .setDefaultProcessName("${ANG_PACKAGE}:bg")
         .build()
 
+    private val applicationScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
+
     /**
      * Initializes the application.
      */
@@ -44,6 +50,9 @@ class AngApplication : Application() {
 
         // Ensure critical preference defaults are present in MMKV early
         SettingsManager.initApp(this)
+        applicationScope.launch {
+            SettingsManager.cleanupOrphanedServerProfilesOnce()
+        }
 
         // Initialize theme state from MMKV
         ThemeManager.refresh()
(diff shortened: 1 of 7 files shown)


2dust:
感谢PR
请分开两个 pr ，一个是清理，一个是重构保存流程。
他们应该分成两次更合理

2dust:
还有个建议，
清理的工作应该独立，不要和所有的业务有关联，清理就是清理，不要在很多地方都调用清理。
因为不是时时刻刻都会产生废弃文件的，不应该在订阅更新或测试等地方又调用清理，这样会导致每个业务的代码耦合在一起。

eliotcougar:
> 还有个建议， 清理的工作应该独立，不要和所有的业务有关联，清理就是清理，不要在很多地方都调用清理。 因为不是时时刻刻都会产生废弃文件的，不应该在订阅更新或测试等地方又调用清理，这样会导致每个业务的代码耦合在一起。

Do you think the orphaned items were the result of failed updates? If so, #6085 should already prevent their accumulation.
If the issue is rare, maybe the cleanup can be made manual? But there is no dedicated "Maintenance" section in the app at the moment.

2dust:
手动清理可能更好也可以接受，因为只有部分人才会产生上万条的垃圾数据。
放到备份和还原中 或设置的高级里面添加一个手动清理，

eliotcougar:
Superseded as requested by two focused draft PRs:

- #6093 — profile save-process hardening
- #6094 — manual profile storage cleanup in Backup & Restore

#6094 has no startup, import, subscription-update, restore, or other automatic cleanup hooks.

PR #6095, opened 2026-08-16
fix(accessibility): label more options buttons

aliRahimi1997:
Add an accessibility label to the More options buttons in Per-app settings and Routing settings.

This allows screen readers to identify the More options actions instead of treating the icons as unlabeled buttons.


2dust:
新添加 string ，请把每个语言文件都补充上，可以用英文

aliRahimi1997:
I checked all the existing locale files, and acc_more is already present in all of them, including the localized strings. So there shouldn't be anything else needed for this string in this PR.

PR #6102, opened 2026-08-17
Optimize probing startup for large subscription groups

eliotcougar:
## Summary

- clear the currently displayed delay values immediately when a group probe starts
- move the per-profile MMKV delay reset from the caller path to the existing I/O coroutine
- complete the reset before starting the test service and reuse the captured GUID list

## Motivation

This is a small optimization extracted from the broader Observatory-based probing proposal in #6050.

Starting a group probe currently calls `clearAllTestDelayResults()` synchronously. That operation decodes, updates, serializes, and writes one MMKV affiliation record per profile, so its caller-path cost grows with the subscription group and is particularly noticeable for groups containing hundreds of profiles.

The change is independent of Observatory probing and benefits both TCPing and RealDelay on the current implementation. It does not change probe selection, concurrency, or result handling.

## Correctness

- Empty groups still return without starting a test.
- Visible results are reset immediately, so old delays are not presented as current while new results arrive.
- Persisted results are reset on `Dispatchers.IO` before `MSG_MEASURE_CONFIG_START` is sent, preventing a late reset from erasing newly reported results.
- The group cache is invalidated after the reset and before the test starts.

## Validation

- `:app:compilePlaystoreDebugKotlin` passes.
- `:app:testPlaystoreDebugUnitTest` runs 46 tests: 44 pass; the existing upstream `UtilsTest.test_isIpAddress` and `UtilsTest.test_IsIpInCidr` failures remain unchanged.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainViewModel.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainViewModel.kt
@@ -689,10 +689,16 @@ class MainViewModel(
         dataSource.cancelAllPing()
         val groupId = uiState.value.selectedGroupId
         val servers = currentServers()
-        dataSource.clearAllTestDelayResults(servers.map { it.guid })
         if (servers.isEmpty()) {
             _uiState.update { it.copy(isTesting = false) }
             return
+        }
+        val serverGuids = servers.map { it.guid }
+        mutableServersForGroup(groupId).update { current ->
+            current.map { server ->
+                if (server.testDelayMillis == 0L) server
+                else server.copy(testDelayMillis = 0L)
+            }
         }
         testingGroupId = groupId
         _uiState.update {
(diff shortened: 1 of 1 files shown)


eliotcougar:
One possible follow-up would be to make the active probe session's in-memory results authoritative while the test is running, instead of using MMKV as the per-result delivery path:

- mark the targeted rows as pending and start probing immediately;
- buffer results by GUID and send the actual `(GUID, delay)` values to the UI;
- clear old persisted results in the background;
- after that clear completes, flush buffered results to MMKV in batches, with a mandatory final flush before auto-sort, invalid-profile removal, or the finish event;
- fence each run with a session ID so late results from a cancelled run cannot affect its replacement.

That would remove the bulk MMKV reset from the probe-start critical path and avoid reloading the entire group from MMKV for every individual result. The clear and flush must remain ordered, otherwise a late clear could erase a newly persisted result.

The Observatory-based probing proposal in #6050 already implements the UI side in a similar way: `CoreTestService` sends a `RealPingResult` containing the GUID and delay, while `MainViewModel` buffers those values in memory, applies them to the cached/visible lists in 500 ms batches, and forces a final flush when probing finishes. It therefore avoids the current per-result full-group reload storm.

#6050 still writes each individual result to MMKV immediately, so deferring and coalescing persistence—and consequently allowing probes to start before the old persistent results have been cleared—would be a possible later extension. This PR deliberately remains the small, conservative optimization: preserve the existing MMKV-authoritative behavior while moving its bulk reset off the UI thread.

PR #6103, opened 2026-08-17
Compose: Locate selected server by stable identifiers

eliotcougar:
## Summary

- represent a pending Locate request with the subscription group ID and server GUID rather than snapshot-dependent group and row indices
- let the existing selected-group state move the pager
- resolve the GUID in the destination page's current list and perform one suspending `scrollToItem`
- remove timing delays, pager-idle coordination, visibility polling, cached index hints, and stale-request bookkeeping

## Why

The previous implementation computed both indices before Compose handled the navigation. Those positions described one snapshot of the UI rather than the server being located.

Servers already have stable GUIDs, and the lazy list and grid already use those GUIDs as item keys. Carrying the same identity through the Locate request avoids positional coupling and lets the destination page resolve its own current index immediately before scrolling.

Locate is a single uninterrupted menu action: the user selects **Find selected configuration** and waits for it to finish. The implementation therefore does not coordinate with concurrent gestures or mutations that are outside this interaction.

## Compose implementation

The ViewModel publishes the target group and GUID as state. The existing selected-group effect moves the pager, while a page-local `LaunchedEffect`, keyed by the target and server list, resolves the GUID and calls the suspending lazy-list or lazy-grid `scrollToItem` API.

This keeps UI work in lifecycle-aware Compose effects and requires no frame delays, polling loops, or imperative registry lookup.

## Performance

Measured on a Pixel 9 Pro API 37 x86_64 emulator with Play Store release builds compiled using `cmd package compile -m speed -f`.

Each run:

1. force-stopped and relaunched the app
2. started from the 15-item `lab` group
3. invoked **More > Find selected configuration**
4. verified that the exact target profile was visible
5. discarded warm-ups and recorded 12 measured trials per build and target

Median results:

| Target | Metric | Upstream | This PR | Change |
| --- | ---: | ---: | ---: | ---: |
| 1,000-item synthetic group, profile 850 | Locate latency | 54.804 ms | 28.368 ms | **-48.2%** |
| | Process CPU | 46.5 ms | 21.0 ms | **-54.8%** |
| 485-item Sub1 group, near-bottom profile | Locate latency | 65.619 ms | 25.345 ms | **-61.4%** |
| | Process CPU | 57.0 ms | 18.0 ms | **-68.4%** |

Median frame and janky-frame counts were effectively unchanged. Upstream produced one background-load outlier for each target; both remain included in the raw samples, while medians are reported so those outliers do not distort the comparison.

## Validation

- `:app:assemblePlaystoreRelease`
- 24 measured emulator Locate runs, all verifying the exact requested target
- full unit suite: 46 tests run; only the two existing unrelated `UtilsTest` environment-dependent failures remained

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/dto/LocateTarget.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/dto/LocateTarget.kt
@@ -2,6 +2,5 @@ package com.v2ray.ang.dto
 
 data class LocateTarget(
     val groupId: String,
-    val groupIndex: Int,
-    val itemPosition: Int
+    val serverGuid: String,
 )
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainContract.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainContract.kt
@@ -66,5 +66,5 @@ sealed interface MainAction {
 
     data class ImportBatchConfig(val configText: String) : MainAction
 
-    data class LocateHandled(val target: LocateTarget) : MainAction
+    data object LocateHandled : MainAction
 }
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainScreen.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainScreen.kt
@@ -30,7 +30,6 @@ import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.v2ray.ang.dto.entities.ProfileItem
 import com.v2ray.ang.ui.compose.LocalDarkTheme
 import com.v2ray.ang.ui.compose.QRCodeDialog
-import com.v2ray.ang.extension.delay
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.launch
 
@@ -73,15 +72,11 @@ fun MainScreen(
     val lazyListStates = remember { mutableStateMapOf<String, LazyListState>() }
     val lazyGridStates = remember { mutableStateMapOf<String, LazyGridState>() }
 
-    var locateInProgress by remember { mutableStateOf(false) }
-
     LaunchedEffect(groups) {
         val validGroupIds = groups.map { it.id }.toSet()
         lazyListStates.keys.retainAll(validGroupIds)
         lazyGridStates.keys.retainAll(validGroupIds)
     }
-
-    val latestDoubleColumnDisplay by rememberUpdatedState(doubleColumnDisplay)
 
     LaunchedEffect(groups, uiState.selectedGroupId) {
         if (groups.isEmpty()) return@LaunchedEffect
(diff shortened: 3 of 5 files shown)


2dust:
你这个修改属于不好优化。
实测会导致滚动变得很慢。

eliotcougar:
> 你这个修改属于不好优化。 实测会导致滚动变得很慢。

I'll look into it.
Is it slow with large subscription groups? I only tested with my 18 configs, and it was instant...

eliotcougar:
I found the cause of the slowdown and pushed the optimization in `7a7d0a77`.

The initial stable-ID implementation did extra work around the scroll:

1. `triggerLocateSelectedServer()` loaded the whole group, scanned it, and published the filtered list.
2. `MainScreen` then dispatched `SelectGroup`, which entered the selected-group loading path and applied the group filtering/list update again. The second load was normally a cache hit, but the repeated O(n) filtering and list work remained.
3. The final GUID-to-index `indexOfFirst` ran in the Compose effect on the UI thread.

This is negligible with a small group, but becomes visible with large subscription groups.

The updated path now:

- loads, filters, and publishes the target group once;
- publishes `selectedGroupId` and `LocateTarget` together, so the screen does not reselect/reload the same group;
- emits a locate target only after the selected GUID is confirmed in the visible list, so missing/failed cases terminate without an arbitrary timeout;
- resolves GUID to index on `Dispatchers.Default`;
- waits for the actual lazy list/grid layout and valid item bounds through `snapshotFlow`, replacing the fixed frame polling;
- suppresses the normal pager synchronization while the locate request owns navigation.

2dust:
在更改类似的功能时，你测试的时候一定要有超过 1000 的数据。
因为在几十个数据的时候，不管什么都会非常快

eliotcougar:
> 在更改类似的功能时，你测试的时候一定要有超过 1000 的数据。 因为在几十个数据的时候，不管什么都会非常快

I know. I wonder where I can find a real subscription group with 1000+ servers to test with... Or I can make a synthetic one, probably. It won't be the same, though. It would be useful to have something with a mix of all kinds of protocols and maybe some invalid configs too.

2dust:
经过考虑，我认为还是用原来的方案更好。
在标准的 android 控件下，都是使用 index；虽然 用 index 有你说的一些问题，但是并不严重。
如果存在非常严重问题，可以针对处理

eliotcougar:
> 经过考虑，我认为还是用原来的方案更好。 在标准的 android 控件下，都是使用 index；虽然 用 index 有你说的一些问题，但是并不严重。 如果存在非常严重问题，可以针对处理

This change is mostly about following Compose guidelines than fixing any issues.

Loaded a synthetic subscription with 1000 socks servers... Locating the server is still instant...
If sorting is enabled, and the selected server is not in the results, it switches to the correct tab, but doesn't scroll, as expected...

2dust:
https://raw.githubusercontent.com/barry-far/V2ray-config/main/Sub1.txt

你用这个数据测试，加倍

eliotcougar:
> https://raw.githubusercontent.com/barry-far/V2ray-config/main/Sub1.txt
> 
> 你用这个数据测试，加倍

Thank you very much.

Let's keep it as a draft for now. I am almost out of tokens at the moment. I'll take another look at it when the limits reset tomorrow. It definitely needs simplification.

One additional thought. Maybe, if the list is filtered, and the selected server is not in the results, locating the server should clear the search instead of failing silently? What's your opinion on that behavior change?

2dust:
> One additional thought. Maybe, if the list is filtered, and the selected server is not in the results, locating the server should clear the search instead of failing silently? What's your opinion on that behavior change?

默认失败比较合理。

eliotcougar:
Maximum optimization... Pure Compose...

PR #6105, opened 2026-08-18
Compose: improve group paging and connection-test results

eliotcougar:
## Summary

- replace the eager group-tab row with a keyed `LazyRow`
- initialize the row around the selected group and scroll only when a new selection is outside the visible range
- emit one connection-test result after the optional endpoint lookup, avoiding two terminal UI updates for one test

## Scope

The service-transition work formerly included here now lives in #6139. Both branches start from upstream `a1b45bbf`. Git merges them cleanly in either order, and both orders produce tree `bea4276e`, exactly matching pre-split #6105 head `af58ee81`.

## Validation

- full `:app:testPlaystoreDebugUnitTest`: 46/46 passed
- `:app:compilePlaystoreDebugKotlin` and x86_64 `:app:assemblePlaystoreDebug`
- API 37.1 / 16 KB emulator: touch and hardware D-pad tab selection, selected-tab persistence and visibility after relaunch, and visual layout
- connection test remained in Testing until one terminal failure result when the endpoint timed out
- crash buffer remained empty

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/core/CoreServiceManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/core/CoreServiceManager.kt
@@ -333,25 +333,14 @@ object CoreServiceManager {
                 }
             }
 
+            val endpoint = if (time >= 0) SpeedtestManager.getRemoteIPInfo() else null
             val result = ConnectionTestResult(
                 delayMillis = time,
                 errorMessage = errorStr,
+                country = endpoint?.country,
+                ipAddress = endpoint?.ipAddress,
             )
             MessageHelper.sendMsg2UI(service, AppConfig.MSG_MEASURE_DELAY_RESULT, result)
-
-            // Only fetch IP info if the delay test was successful
-            if (time >= 0) {
-                SpeedtestManager.getRemoteIPInfo()?.let { ip ->
-                    MessageHelper.sendMsg2UI(
-                        service,
-                        AppConfig.MSG_MEASURE_DELAY_RESULT,
-                        result.copy(
-                            country = ip.country,
-                            ipAddress = ip.ipAddress,
-                        ),
-                    )
-                }
-            }
         }
     }
 
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainGroupTab.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainGroupTab.kt
@@ -1,19 +1,27 @@
 package com.v2ray.ang.ui.main
 
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
+import androidx.compose.foundation.layout.widthIn
+import androidx.compose.foundation.lazy.LazyRow
+import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.PrimaryScrollableTabRow
 import androidx.compose.material3.Tab
-import androidx.compose.material3.TabRowDefaults
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.text.style.TextOverflow
-import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.v2ray.ang.dto.GroupMapItem
@@ -28,32 +36,34 @@ fun GroupTabBar(
     onTabClick: (Int) -> Unit,
     modifier: Modifier = Modifier
 ) {
-    PrimaryScrollableTabRow(
-        selectedTabIndex = selectedTabIndex.coerceIn(0, groups.lastIndex),
-        modifier = modifier.fillMaxWidth(),
-        containerColor = MaterialTheme.colorScheme.surface.copy(alpha = 0.95f),
-        contentColor = MaterialTheme.colorScheme.onSurface,
-        edgePadding = 16.dp,
-        minTabWidth = 56.dp,
-        indicator = {
-            TabRowDefaults.PrimaryIndicator(
-                modifier = Modifier
-                    .tabIndicatorOffset(
-                        selectedTabIndex = selectedTabIndex.coerceIn(0, groups.lastIndex),
-                        matchContentSize = true
-                    )
-                    .clip(RoundedCornerShape(3.dp)),
-                width = Dp.Unspecified,
-                color = MaterialTheme.colorScheme.secondary
-            )
-        },
-        divider = {}
+    val selectedIndex = selectedTabIndex.coerceIn(0, groups.lastIndex)
+    val listState = rememberLazyListState(initialFirstVisibleItemIndex = selectedIndex)
+
+    LaunchedEffect(selectedIndex) {
+        if (listState.layoutInfo.visibleItemsInfo.none { it.index == selectedIndex }) {
+            listState.animateScrollToItem(selectedIndex)
+        }
+    }
+
+    LazyRow(
+        state = listState,
+        modifier = modifier
+            .fillMaxWidth()
+            .background(MaterialTheme.colorScheme.surface.copy(alpha = 0.95f)),
+        contentPadding = PaddingValues(horizontal = 16.dp),
+        verticalAlignment = Alignment.CenterVertically,
     ) {
-        groups.forEachIndexed { index, group ->
+        itemsIndexed(
+            items = groups,
+            key = { _, group -> group.id },
+        ) { index, group ->
+            val serverFlow = remember(group.id, mainViewModel) {
+                mainViewModel.serversForGroup(group.id)
+            }
             GroupTabItem(
                 group = group,
-                selected = index == selectedTabIndex,
-                serverFlowProvider = { mainViewModel.serversForGroup(group.id) },
+                selected = index == selectedIndex,
+                serverFlow = serverFlow,
                 onClick = { onTabClick(index) }
             )
         }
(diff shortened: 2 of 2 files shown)


aliRahimi1997:
As a blind TalkBack user, I tested these changes on a real device.

The new selection semantics work technically, but I noticed some announcement-order issues that make the experience less comfortable for screen reader users.

In the server list, unselected servers are announced with “Not selected” before the server information. Since most items in the list are unselected, this adds repetitive information before the content the user actually needs to identify the server.

In the app list, the behavior is better when an item is unchecked: the app name is announced first, followed by its state. However, after checking an item, TalkBack announces the checked state before the app information and then additionally announces that it is a checkbox. This changes the announcement order and makes the result less consistent.

I think it may be worth considering whether the selection/toggle semantics can be exposed without putting the state before the item's main information, while still preserving the correct semantic role for accessibility.

Thank you for working on improving accessibility in the app. I wanted to share this from the perspective of an actual blind TalkBack user.

eliotcougar:
> As a blind TalkBack user, I tested these changes on a real device.
> 
> The new selection semantics work technically, but I noticed some announcement-order issues that make the experience less comfortable for screen reader users.
> 
> In the server list, unselected servers are announced with “Not selected” before the server information. Since most items in the list are unselected, this adds repetitive information before the content the user actually needs to identify the server.

Got it. Does it announce all unselected items or only those currently on screen? I imagine you need to know what servers are currently on screen, unselected, and available to be selected, then you can tap and hear what server you tapped, and then you confirm to select it. What is the preferred workflow? Drop the useless phase "Not selected" that wastes time?

> In the app list, the behavior is better when an item is unchecked: the app name is announced first, followed by its state. However, after checking an item, TalkBack announces the checked state before the app information and then additionally announces that it is a checkbox. This changes the announcement order and makes the result less consistent.

This should be addressed by #6105 at least partially - elements had a duplicate checkbox element as a leftover from Compose migration.

> I think it may be worth considering whether the selection/toggle semantics can be exposed without putting the state before the item's main information, while still preserving the correct semantic role for accessibility.

We will look into decoupling the visual layout from the semantic content.

> Thank you for working on improving accessibility in the app. I wanted to share this from the perspective of an actual blind TalkBack user.

Thank you for testing and feedback. It is extremely valuable.

UPD: Just realized I can simply enable Talkback and test. D-uh...

aliRahimi1997:
Yes. TalkBack announces “Not selected” for every unselected server item that I navigate to in the list. It is not limited to the currently selected server or to a selection change.

One clarification about the server list: a server cannot really be “unselected” independently. Selecting another server is what changes the previous server from selected to unselected.

Because of that, I don't think “Not selected” provides much useful information for the other servers.

My preferred workflow would be:

- Unselected server: server information only.
- Selected server: “Selected” first, followed by the server information.
- After selecting another server: the newly selected server should be announced with “Selected” so the user immediately knows that the selection changed.

For example:

Selected → server information

or:

server information

for an unselected server.

This way, a TalkBack user can immediately identify the active server when focused on it, without having to hear “Not selected” repeatedly for every other server in the list.

2dust:
因为我们无法测试获得效果，请你们使用者反馈。
准备可以合并了通知

eliotcougar:
Updated the TalkBack accessibility behavior in the focused branch.

The original `selectable(selected = false, role = Role.RadioButton)` semantics published an explicit unselected state, which caused TalkBack to announce “Not selected” for every non-current server. Server rows now retain the radio role and click action, but expose the selection state only for the currently selected row. The selected row uses the localized “Currently selected server” state description before the rest of its content.

Delay values now keep their abbreviated visual units (`ms`, `мс`, etc.). A separate localized accessibility resource is applied through the delay Text’s `contentDescription`, so TalkBack receives full unit names such as “milliseconds” while the UI remains compact.

Validation completed:
- Focused branch Kotlin compilation passed.
- Hands-on tested with TalkBack on.

PS: When the server is made active there is a flood of audible notifications, including some duplicate ones. I will try to address that in a separate PR.

eliotcougar:
It now fully overwrites #6108 with fully localized spoken phrase for the selected server and no phrase for an unselected server.

eliotcougar:
@aliRahimi1997 concerns over ServerList items Talkback announcements have been addressed.
In my testing, the blind navigation feels much more natural now on that specific screen. Other accessibility concerns will be addressed separately when I get to them.

@2dust This PR can be merged as is.

aliRahimi1997:
I tested the latest version with TalkBack on a real device. The unselected-server behavior is now excellent — “Not selected” is no longer announced. Thank you!
I only noticed two small things in the selected server announcement. It currently says “Currently selected server” before the server information, and “Radio button” is announced at the end.
Would it be possible to shorten the localized state description to simply “Selected”, and avoid announcing “Radio button” at the end if possible? Other than these two small details, the current behavior works very well for me.

eliotcougar:
> I tested the latest version with TalkBack on a real device. The unselected-server behavior is now excellent — “Not selected” is no longer announced. Thank you! I only noticed two small things in the selected server announcement. It currently says “Currently selected server” before the server information, and “Radio button” is announced at the end. Would it be possible to shorten the localized state description to simply “Selected”, and avoid announcing “Radio button” at the end if possible? Other than these two small details, the current behavior works very well for me.

Hmmm... I tried to make it intentionally explicit. The text in the server's name can be weird, and I think it is better to have it say "Currently selected server - a bunch of nonsense" instead of just "Selected - a bunch of nonsense".

In case of the "Radio button"/"Checkbox"/"Tab"/"Button" at the end, I'm not sure it's configurable... Talkback always announces the type of control at the end. I'll check the docs now...

aliRahimi1997:
I understand your reasoning for using “Currently selected server”, especially when server names can contain arbitrary or confusing text.
However, in my original #6108 implementation, the state description was simply localized as “Selected”, which was shorter and, in my real-device TalkBack testing, worked very naturally.
Also, my original implementation used regular clickable(onClick = onClick) without Role.RadioButton, so TalkBack did not announce “Radio button”.
So from my testing, I actually preferred the original behavior:
“Selected” → server information
without the additional “Radio button” announcement.
If possible, I would prefer keeping that behavior.

aliRahimi1997:
One more accessibility issue I noticed in the app:
There is a setting that, when enabled, asks for confirmation before deleting a server/profile. However, the confirmation dialog currently uses a generic message and does not announce the name of the profile/server that is about to be deleted.
For TalkBack users, this makes it difficult to know exactly which profile is being deleted, especially when there are multiple profiles with similar names.
Could this confirmation dialog also include the name of the profile/server being deleted, if possible?
Thank you!

eliotcougar:
Updated the focused branch with the latest accessibility changes.

- Removed the server-row radio-button role and selected-state semantics.
- Active server rows now expose merge-only semantics with no click action, so TalkBack should not announce double-tap activation for the already-selected server.
- Inactive server rows remain clickable.
- The localized “Selected server” prefix remains attached to the active row’s merged content and is spoken before the server details.
- Delete confirmations now include the profile name after the question. This is applied to the main server list and profile editor delete dialogs across all maintained locales.

aliRahimi1997:
Thanks for the update. I tested the latest commit on a real device with both TalkBack and Commentary.
The improvement for unselected servers is excellent — TalkBack no longer announces “Not selected”.
However, I noticed an important regression compared with my previous implementation. In my implementation, the server row kept the same accessible content for both selected and unselected servers, and I did not have this issue with either TalkBack or Commentary.
With the current implementation, Commentary behaves differently: for an unselected server it reads the full server information, but for the selected server it only reads the server name and does not expose the rest of the server information or its selected state.
I also noticed that TalkBack does not announce the selected state immediately after selecting a server. I have to move the focus away and back to the server before “Selected server” is announced.
So I wonder if we could keep the approach from my original implementation for the server-row content, and only change the semantics needed to prevent “Not selected” from being announced for unselected rows.
My main concern is that the accessibility behavior should remain consistent across different screen readers. The previous implementation did not have this Commentary regression, while the current approach seems to solve the TalkBack issue at the cost of changing what other screen readers can access.
Apart from this, I think the current behavior is very close to what we need.

eliotcougar:
> Thanks for the update. I tested the latest commit on a real device with both TalkBack and Commentary. The improvement for unselected servers is excellent — TalkBack no longer announces “Not selected”. However, I noticed an important regression compared with my previous implementation. In my implementation, the server row kept the same accessible content for both selected and unselected servers, and I did not have this issue with either TalkBack or Commentary. With the current implementation, Commentary behaves differently: for an unselected server it reads the full server information, but for the selected server it only reads the server name and does not expose the rest of the server information or its selected state. I also noticed that TalkBack does not announce the selected state immediately after selecting a server. I have to move the focus away and back to the server before “Selected server” is announced. So I wonder if we could keep the approach from my original implementation for the server-row content, and only change the semantics needed to prevent “Not selected” from being announced for unselected rows. My main concern is that the accessibility behavior should remain consistent across different screen readers. The previous implementation did not have this Commentary regression, while the current approach seems to solve the TalkBack issue at the cost of changing what other screen readers can access. Apart from this, I think the current behavior is very close to what we need.

Supporting other screen readers complicates things a lot. GPT5.6 did warn me about it. A lot of customization and workarounds in Compose accessibility primarily targets TalkBack.

Because I was out of AI tokens today, I had to work with the cheap GPT5.6-Luna. It is so dumb and annoying to work with. I'll set the PR to draft for now. Tomorrow, when I get access to Sol again, I can do a proper research into interoperability with third-party screen-readers.

Meanwhile, there is a present for you in the form of #6112.
It is late, and I have important work tomorrow at the lab, so I will disappear for the most of the day.

eliotcougar:
The main screen accessibility semantics are fully functional, but it's not pretty yet under the hood. Let me do a couple more cleanup passes before it's ready for merge.

eliotcougar:
The two PRs (this and #6112) should have no intersections and can be merged in any order, hopefully. And you can test them together if you want.

I think I found all of the most obvious quirks. I can't do anything with some out of order messages when service is started/stopped/restarted, as a lot of events happen asynchronously, and the notifications arrive whenever they want.

There is a problem I don't yet know how to approach. Persian and Bakhtiari don't have TTS engines on Android. If you select that locale in the app and enable TalkBack - no sound. I imagine someone may want to set the UI to Persian, but use English for TalkBack, for example.

The toast notifications rework can potentially be separated into an isolated PR. It's just a lot of single-line changes throughout the entire codebase where toasts are called, to make them either silent or audible, depending on the situation.

As always, the normal app behavior is almost unaffected.

The final part of the accessibility work (all the editing screens) can be done after these.

2dust:
请问是否你们搞得过度了。
请要遵循 google 官方的方案，因为后面我们是让人来维护的

V2rayNG/app/src/main/java/com/v2ray/ang/service/CoreVpnService.kt
```diff
@@ -87,6 +89,9 @@ class CoreVpnService : VpnService(), ServiceControl {
             return START_NOT_STICKY
         }
         LogUtil.i(AppConfig.TAG, "StartCore-VPN: Service command received, systemVpnStart=$isSystemVpnStart")
+        if (isSystemVpnStart) {
+            toast(R.string.toast_services_start)
```
2dust: 这里用 toast 是不可接受的

eliotcougar: Reverted

eliotcougar:
> 请问是否你们搞得过度了。 请要遵循 google 官方的方案，因为后面我们是让人来维护的

I agree... This whole accessibility endeavor encouraged by @aliRahimi1997 resulted in working, but disorganized accessibility code.

I can move all the accessibility stuff to #6112 and mark it as a long-term draft until I learn how best to adapt the app for screen readers, so that the Compose part of this PR can be merged.

eliotcougar:
Alright, we have reduced the scope and addressed the edge cases in service state flow indication. The accessibility work is on hold for now in a separate PR.

eliotcougar:
Refined the restart lifecycle in af58ee81:

- Replaced the anonymous restart coroutine with a named, CoreServiceManager-owned job lifecycle.
- Explicit stop now cancels pending restart work; duplicate and stale restart operations are rejected.
- Added focused tests for ownership, cancellation, duplicate requests, and stale tokens.

Validation passed: focused and full Play Store debug unit tests, Kotlin compilation, x86_64 APK assembly, and API 37.1 start/restart/stop checks in VPN and proxy-only modes. The crash buffer remained empty. Root mode was not available on this emulator because it has no su binary.

eliotcougar:
The service-transition changes have moved to #6139. This PR now contains only the focused Compose group-paging work and the single-final-update connection-test change.

Both branches remain based on `a1b45bbf`. I verified them with Git in both merge orders; each produces tree `bea4276e`, exactly matching the pre-split #6105 tree at `af58ee81`.

PR #6108, opened 2026-08-18
fix(accessibility): announce selected server state

aliRahimi1997:
## Accessibility improvement

Expose the selected server state to screen readers.

### Changes
- Expose the selected server state through Compose `stateDescription`.
- Add localized accessibility strings.
- Avoid announcing "Not selected" for unselected servers.
- Keep the existing clickable and accessibility behavior unchanged.


2dust:
新添加 string ，请把每个语言文件都补充上，可以用英文

PR #6109, opened 2026-08-18
Use native Xray TUN for connection info and updates

### LINKED ISSUE #5523: Country and IP of the server after the connection test missing with dynamic port, user and password
I recently installed version 2.1.1 and noticed that when I test the connection using the button at the bottom of the program, the server's country and IP are no longer shown like in version 1. How can I enable this? I only see ping. The URL for information in the settings opens normally in the browser.

eliotcougar:
## Summary

- restore country and IP enrichment for current-connection tests in native Xray TUN mode when the local proxy is disabled
- download subscriptions, application-update metadata, and asset files through the already-running Xray core in the same mode
- force ordinary profiles through the `proxy` outbound
- resolve a primary policy group's first viable principle target and use that concrete member
- preserve the existing local-proxy and direct fallback paths for other modes, including HEV and root configurations that force a local proxy
- preserve request headers, including custom `Host` authorities, user agents, and embedded URL credentials
- prevent incomplete or overlapping asset downloads from replacing valid files
- keep daemon URL downloads and service controls private to this app
- cancel pending daemon downloads with the core service lifecycle

## Root cause

The connection test measures delay and then requests **Current connection info test URL** to display the exit country code and IP address. Subscription and asset updates similarly relied on an HTTP client configured for the local proxy. Native Xray TUN can run with that proxy disabled, leaving no listener for these requests even though the core and tunnel are active.

Application update checks had a related problem: the app first fetched GitHub release metadata directly. Because v2rayNG excludes its own process from its VPN interface, that request could bypass the active Xray connection. With the local proxy disabled, there was no app-owned path that followed the selected outbound.

An unqualified core request would not be equivalent: user routing rules could send the URL or its resolved IP through `direct` or another detour, producing incorrect connection information or bypassing the selected route for updates.

The daemon also registered its internal service action in the same dynamic receiver filter as system screen broadcasts. On Android 13 and newer, that mixed receiver was exported, allowing another app to send internal control messages. Once `MSG_DOWNLOAD_URL` was added, this included arbitrary URL, header, timeout, and `ResultReceiver` input, creating a request and success oracle through the user's active tunnel.

## Implementation

The app sends eligible requests to the daemon process, which downloads them through the running core and an explicit outbound. Ordinary profiles use `proxy`; primary policy groups resolve `balancer-main` and use its first non-empty principle target.

One shared eligibility decision now controls connection-info and app-owned download routing. It selects the core only for native Xray TUN with the local proxy disabled, and excludes HEV, root, and root-LAN configurations because those modes force a local proxy regardless of the visible setting.

The application update checker uses the same core download bridge for GitHub release metadata. Once the daemon accepts an eligible request, a transfer failure is reported instead of silently retrying outside the tunnel. The eventual APK link still opens in the user's browser, whose traffic is handled by the VPN normally.

Subscription headers, custom user agents, custom `Host` authorities, and embedded URL credentials are preserved. Binary and large responses are streamed through a UUID-constrained cache file, avoiding string corruption and Binder payload limits.

Asset updates use a separate UUID staging file for every attempt. This prevents overlapping updates from writing through the same temporary inode. A downloaded asset is promoted only when it is non-empty; direct HTTP downloads also reject unsolicited partial-content responses and declared-length mismatches. The AndroidLib method writes through a sibling temporary file and atomically replaces its destination only after the transfer completes, so network, write, or close failures preserve the previous file.

The internal service-action receiver is separate from the screen-state receiver and registered with `RECEIVER_NOT_EXPORTED`. System screen broadcasts keep their existing receiver behavior. URL downloads run in a service-owned `SupervisorJob` scope; stopping the core cancels the scope and prevents a cancelled request from delivering a late `ResultReceiver` callback. The conflict resolution retains upstream request IDs, cancellation callbacks, and the dedicated connection-test scope while keeping URL downloads independently owned by the core lifecycle.

Fallback behavior remains deliberate:

- native Xray TUN with local proxy disabled: use the running core
- daemon unavailable: use the existing direct fallback without first waiting on the disabled proxy
- accepted core request fails: report failure instead of silently bypassing the tunnel
- local proxy enabled, HEV TUN, root mode, root-LAN sharing, and proxy-only mode: retain their existing paths

For strategies that expose several principle candidates, policy-group selection is intentionally best-effort and does not advance random or round-robin state.

## AndroidLib dependency

This branch updates the AndroidLibXrayLite gitlink to [2dust/AndroidLibXrayLite#205](https://github.com/2dust/AndroidLibXrayLite/pull/205), which provides:

- `CoreController.GetUrlContent(url, outboundTag)`
- `CoreController.GetBalancerPrincipleTarget(balancerTag)`
- `CoreController.DownloadUrlToFile(url, outboundTag, headersJSON, filePath, timeoutMillis)`
- correct Go handling for custom `Host` authorities
- transactional file replacement and response-length validation for downloads

## Validation

- AndroidLib `go vet ./...`
- AndroidLib `go test -race ./...`
- four-ABI AndroidLib AAR build and generated API/ABI inspection
- all 94 Play Store debug JVM tests after rebasing onto upstream connection-test lifecycle handling
- Play Store debug Kotlin compilation and APK assembly
- produced APK inspection for `libgojni.so`, `libhev-socks5-tunnel.so`, and `libhevsockstun.so` in `armeabi-v7a`, `arm64-v8a`, `x86`, and `x86_64`
- clean x86_64 emulator feature validation with native Xray TUN, HEV disabled, and local proxy disabled
- subscription update and three asset downloads completed through the active core without a loopback proxy connection or crash
- API 33 sender-boundary check: an external shell-UID service-control broadcast was not delivered and the tunnel remained connected
- same-app UI service-control broadcast still stopped the service normally

Not run after the latest update-check change: an end-to-end application update check through a live native Xray TUN session, because no emulator or physical device was attached.

Android lint still reports four pre-existing `NewApi` errors in untouched root/theme files; it reports no errors in this PR's changed files.

Fixes #5523


2dust:
关闭了本地端口的用户，只是实现测试连接还是不够的；因为更新订阅和下载geo 等功能也是要依赖这个本地端口。
所以需要找到能解决上面所有问题的方法，不然只实现测试不够

eliotcougar:
> 关闭了本地端口的用户，只是实现测试连接还是不够的；因为更新订阅和下载geo 等功能也是要依赖这个本地端口。 所以需要找到能解决上面所有问题的方法，不然只实现测试不够

In case of Assets and Subscriptions, should we let users choose "Update without proxy" vs "Update through proxy", like in v2rayN?..

eliotcougar:
Security follow-up pushed in a00589a4.

The daemon's dynamic receiver previously combined `BROADCAST_ACTION_SERVICE` with `ACTION_SCREEN_ON/OFF` and used the exported receiver flag on Android 13+. Setting the destination package limited where the intent went, but did not authenticate who sent it. After `MSG_DOWNLOAD_URL` was introduced, another app could therefore supply an arbitrary URL, headers, timeout, and `ResultReceiver`, using the active Xray connection as a request/success oracle. The same receiver also exposed the other internal service-control message keys.

The fix separates the two trust boundaries:

- `BROADCAST_ACTION_SERVICE` is registered on an app-private `RECEIVER_NOT_EXPORTED` receiver.
- System screen-state broadcasts retain a separate receiver with the existing flags.
- URL downloads use a service-owned `SupervisorJob + Dispatchers.IO` scope.
- Core shutdown cancels pending downloads, and cancellation is checked before sending the `ResultReceiver` callback.

Validation on an API 33 emulator confirmed that an external shell-UID stop message was not delivered and the tunnel remained connected, while the app's own UI stop message still worked. The focused unit test and Play Store debug build also pass.

@2dust, a related but broader privacy issue is that existing code still writes some complete subscription/configuration/download/WebDAV URLs, and some URL-bearing exception messages, to Logcat. Android's guidance recommends removing sensitive values and keeping production log messages predictable: https://developer.android.com/privacy-and-security/risks/log-info-disclosure

Would you be open to a separate, small PR that removes those URL values and retains only operation names and safe status codes? I kept that cleanup out of #6109 to avoid expanding this PR's scope and merge noise.

TiAmo-jy (approved):

PR #6097, opened 2026-08-16
Add translator credits page

eliotcougar:
## Summary

- add an About > Translators page for #6070
- credit major contributors for each of the eight maintained non-English locales
- show each language's localized name together with its English name
- link GitHub profiles that still resolve while retaining historical names whose accounts no longer resolve
- include the requested note that the recent all-locale localization audit was carried out with GPT-5.6-Sol and reviewed by `eliotcougar`
- localize the page copy across all nine maintained string catalogs

English is intentionally excluded from the translator list.

Closes #6070

## How the lists were compiled

The audit started from `upstream/master` at `e4e3a36e` and examined the history of every maintained `values*/strings.xml` catalog. For each non-English locale, I used no-merge Git history and per-path shortlogs to identify authors who changed that catalog, then inspected the relevant commits to distinguish translation-focused work from feature additions, mechanical synchronization, accessibility work, and other resource-file changes.

Contributor aliases were normalized using GitHub account identity and commit email where possible. Two historical names were confirmed as renamed accounts: `@MH140000` is now `@mh292929`, and `@NagisaEfi` is now `@Fubuki0x10DE`. Every profile linked from the app was checked directly. Historical handles that no longer resolve are kept as visible, non-clickable credits rather than silently discarded.

The compact lists shown in the app are the locale-focused major contributors. The broader audit below intentionally contains identifiable Git authors who touched each locale catalog, except contributors who opted out; it is included for review and provenance, but those names should not all be interpreted as translators. The GPT-5.6-Sol audit note is based on the recent issue/PR localization work and is non-Git attribution.

## Credits shown in the app

- Arabic: `@MrIbrahem`
- Bengali: `@CodeWithTamim`
- Luri Bakhtiari: `@hosseinabaspanah`, `@CodeWithTamim`
- Persian: `@TheMRVX`, `@Skh-web6982`, `@Ptechgithub`, `@Pk-web6936`, `@alphax-hue3682`, `@phoenix6936`, `@DecorativeFamily`, `@decorativeman`, `@mh292929`, `@Amir-yazdanmanesh`, `@CUMOON`, `@hadi-norouzi`, Vahid Farid
- Russian: `@solokot`, `@Liniya`, `@eliotcougar`
- Vietnamese: `@admarty`, `@user09283`, `@yuhan6665`
- Simplified Chinese: `@Yau08`
- Traditional Chinese: `@Yau08`, `@Fubuki0x10DE`

## Expanded catalog-history audit

<details>
[...]

PR #6114, opened 2026-08-20
Add Translation Contributor Page

codewithtamim:
cc @2dust
image


2dust:
https://github.com/2dust/v2rayNG/pull/6097
需要接受哪个更合适？

codewithtamim:
> #6097 需要接受哪个更合适？

#6114

eliotcougar:
I don't know... The benefit from storing translators as json is minimal, and it adds a whole parsing step. It would make sense if the json was fetched from the Internet.

codewithtamim:
> I don't know... The benefit from storing translators as json is minimal, and it adds a whole parsing step. It would make sense if the json was fetched from the Internet.

I dont think that hardcoding that way is good. As you said "It would make sense if the json was fetched from the Internet."

The app already has v2ray_config.json and  v2ray_config_with_tun.json and parses them at runtime. So in your logic we need to change those also :) Cause they are not fetched form the internet either. If parsing the json would be a problem then those would be hardcoded into lets say kotlin maps too :0

the benifit is that that contributors data is not code anymore, we put them into assets as json and parse them runtime like the way they are done.

the whole point is to keep it organized/consistent and in a way so devs can pick up quick.

eliotcougar:
> The app already has v2ray_config.json and v2ray_config_with_tun.json and parses them at runtime. So in your logic we need to change those also :) Cause they are not fetched form the internet either. If parsing the json would be a problem then those would be hardcoded into lets say kotlin maps too :0

Condescending attitude is unjustified here. v2ray configs are json because we supply them to xray-core as json.

Parsing json is not a problem. Editing TranslatorsActivity.ks is also not a problem. Fetching kotlin from the Internet is a problem.

In any case, from my point of view both approaches have no advantages or disadvantages over one another.

codewithtamim:
> > The app already has v2ray_config.json and v2ray_config_with_tun.json and parses them at runtime. So in your logic we need to change those also :) Cause they are not fetched form the internet either. If parsing the json would be a problem then those would be hardcoded into lets say kotlin maps too :0
> 
> Condescending attitude is unjustified here. v2ray configs are json because we supply them to xray-core as json.
> 
> Parsing json is not a problem. Editing TranslatorsActivity.ks is also not a problem. Fetching kotlin from the Internet is a problem.
> 
> In any case, from my point of view both approaches have no advantages or disadvantages over one another.



> > The app already has v2ray_config.json and v2ray_config_with_tun.json and parses them at runtime. So in your logic we need to change those also :) Cause they are not fetched form the internet either. If parsing the json would be a problem then those would be hardcoded into lets say kotlin maps too :0
> 
> Condescending attitude is unjustified here. v2ray configs are json because we supply them to xray-core as json.
> 
> Parsing json is not a problem. Editing TranslatorsActivity.ks is also not a problem. Fetching kotlin from the Internet is a problem.
> 
> In any case, from my point of view both approaches have no advantages or disadvantages over one another.

Exactly, — You are — right

hosseinabaspanah:
Friends @eliotcougar and @codewithtamim  , all this disagreement and discussion about implementing this feature only breeds resentment.

It's better for @2dust to make the final decision on which method is better.

Please be kind.😃

2dust:
不用争论，用 json 确实稍微好一些。
感谢大家的贡献

PR #6120, opened 2026-08-22
Improve service controls for Android TV

Lostonelzq:
## Summary

- detect television devices using the masked UI mode type or `FEATURE_LEANBACK`
- place the start/stop service action in the top app bar on TV devices so it is reachable with a D-pad
- keep the existing floating action button unchanged on non-TV devices

## Motivation

On some Android TV devices and projectors, the floating service button cannot be reached using the remote control. Moving the same action next to the add button for TV layouts makes service control accessible without introducing a separate UI.

## Testing

- built and installed the F-Droid variant on a JMGO N1 Pro projector
- verified D-pad focus and start/stop behavior
- verified the non-TV floating action button remains selected by the existing layout path

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainBottomBar.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainBottomBar.kt
@@ -39,6 +39,7 @@ fun MainBottomBar(
     displayText: String,
     isRunning: Boolean,
     isDarkTheme: Boolean,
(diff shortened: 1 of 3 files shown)


eliotcougar:
Full TV UI already exists locally. It just waits for the Compose migration to be finished.
https://github.com/eliotcougar/v2rayNG/tree/google-tv-ui


Interim partial updates are potentially useful if the app is completely unusable in its current state.

Lostonelzq:
Okay, understood. Thanks for the clarification.

PR #6121, opened 2026-08-22
Fix #6117  Provide Android Developer Key Hash 

### LINKED ISSUE #6117: Publish the android development key fingerprint
Publish the Android key (along with the pgp key) to use with AppVerifier. 

Please verify whether the following extracted key is legitimate.  
```
com.v2ray.ang
B7:DC:88:3E:76:78:D6:C6:18:56:9E:48:A1:C8:C2:67:A1:34:A7:5F:8B:CB:76:35:10:C8:BF:FB:3B:90:B0:77
```

expendeducated:
Used for app verification

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,15 @@ Release files are signed with GPG to verify authenticity and integrity, helping 
 
 ---
 
+## Android Developer Key Hash / 安卓开发者公钥指纹
+
+```text
+com.v2ray.ang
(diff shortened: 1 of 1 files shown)


expendeducated:
The maintainer refused to document their Android development public key. This refusal puts the users at significant risk of downloading a malicious copy resulted from a man-in-the-middle attack. I advice caution to all users of this software.

PR #6125, opened 2026-08-22
Isolate and optimize RealDelay probing without library changes

### LINKED ISSUE #6089: Delay test UI bug

### LINKED ISSUE #6143: 延迟测试后，应该关闭这个连接
最新版本
使用最小面点击延迟测试后，这个连接会一直连着，不会立马关闭。
我觉得我们得到延迟信息后，我们马上可以关闭这个连接了。

eliotcougar:
## Context

This is a focused, app-only follow-up to #6050 based on the maintainer feedback there. It keeps the existing per-profile RealDelay implementation, does not use Xray Observatory, and makes no changes to `AndroidLibXrayLite` or xray-core.

The goal is to address the isolation, concurrency, progress, cancellation, and UI-delivery problems without introducing a batch-config builder or new native APIs.

Fixes #6143
Fixes #6089

## What changes

- Run interactive probes in one shared disposable `:Probe` service process instead of the process that hosts the long-running VPN core.
- Run subscription-update probes in their own `:SubscriptionUpdate` process for the same reason.
- Keep one existing short-lived Xray core per profile, but apply the configured concurrency limit to the actual worker dispatcher. Custom configurations are additionally serialized because their full native configs can interfere during teardown.
- Skip policy-group rows for RealDelay and TCPing. Cold-starting a policy group can select the wrong member; support can be revisited after #6109 exposes the target selected by the running core. Interactive probing uses the already-materialized row type, so this adds no serial profile-decode pass before scheduling.
- Carry the test session ID, profile GUID, and delay through the result pipeline.
- Coalesce keyed row updates every 500 ms, force the final pending flush before completion, and preserve unchanged row objects. The UI no longer invalidates, reloads, and republishes the complete group for every result.
- Throttle progress delivery separately to 100 ms. Progress counts completed scheduled probes, while successful completion ends with `x/y servers live` for the probeable rows.
- Make an active batch cancellable from the foreground notification and reject stale events from replaced/cancelled sessions.
- Apply the same coalesced result path to TCPing as well as RealDelay.

## Difference from #6050

| | #6050 Observatory proposal | This PR |
|---|---|---|
| Native/library changes | New AndroidLib APIs and a modified native library | **None**; the AndroidLib gitlink and tested AAR are byte-identical to upstream |
| Probe model | Combined compatible profiles into one Observatory core, with fallbacks | Retains individual per-profile probes; policy rows are skipped until the running-core target API from #6109 is available |
| Xray Observatory | Required | **Not used** |
| Process safety | Disposable probe process | Disposable probe process |
| UI delivery | Identified, coalesced results | Identified, coalesced results, without the batch-probing machinery |
| Maintenance surface | App plus AndroidLib/native adapter and combined-config logic | App-side service lifecycle, event DTOs, and UI state only |

The Observatory approach remains faster for its natural one-request-per-profile workload. This version deliberately gives up that optimization to keep the native library unchanged and the review/maintenance boundary smaller.

## Related bugs

This should address the same user-visible problems cited by #6050: #6077 and probably #6089.

For #6089 specifically, the permanent OEM/device-specific freeze was not reproduced on the emulator, so this PR does not claim controlled proof of the root cause. It does remove the two leading app-side contributors: lossy/bursty identity-less result delivery and the per-result full-group reload/recomposition queue. It also covers the reported TCPing path.

## 500-profile performance comparison

Controlled emulator workload:

- one Android 13/API 33 x86_64 emulator, four virtual CPUs;
- one deterministic subscription containing 500 SOCKS profiles;
- emulator-local SOCKS and HTTP fixtures, with each HTTP response delayed by 100 ms;
- configured concurrency 16;
- one excluded warm-up and five accepted measured runs per build;
- CPU sampled every 250 ms for every app process; memory sampled approximately once per second;
- upstream `63f557242bdd071214c4037c76c912b66da925c8` (current when the controlled set was collected) versus candidate `1408ef43d753e8fca54acf4e92563f8af1a02965`;
- the same byte-identical upstream AAR in both builds.

The final branch is rebased onto current upstream `13138dddcab291ba3618cc54721ce4c7dd926be0`. The policy-skip follow-up does not change the worker loop, and this 500-profile dataset contains no policy groups; the table reports the retained controlled five-run set rather than mixing in later runs from an emulator session that developed abnormal QEMU disk activity.

Values below are medians; ranges and raw samples were retained locally.

| Metric | Upstream baseline | This PR | Change |
|---|---:|---:|---:|
| Network complete | 18.06 s | 13.08 s | **27.6% faster** |
| Final UI observed | 20.87 s | 15.93 s | **23.7% sooner** |
| Total app CPU | 35.57 s | 17.94 s | **49.6% less** |
| Main-process CPU | 18.13 s | 5.38 s | **70.3% less** |
| Average app CPU | 170.3% of one core | 114.5% | **32.7% lower** |
| Peak combined PSS | 252.7 MiB | 284.7 MiB | **12.7% higher** |
| Peak summed RSS | 405.7 MiB | 513.1 MiB | **26.5% higher** |
| Frames rendered during flow | 295 | 215 | **27.1% fewer** |
| Janky frames | 170 | 115 | **32.4% fewer** |
| Frame p50 / p99 | 48 / 550 ms | 32 / 250 ms | improved |

The CPU reduction includes all app processes. The memory increase is the expected tradeoff of process isolation: during the measured peak, the main process, normal daemon, and disposable `:Probe` process overlapped. The probe process was gone after every completed run.

Every accepted upstream and candidate run completed the full natural workload: 1,000/1,000 HTTP GETs, 1,500/1,500 socket accepts, and exact observed HTTP/SOCKS concurrency of 16. No crash, ANR, SIGSEGV, or native abort marker was found. The gfxinfo values are directional emulator/debug-build evidence rather than production frame-rate claims.

### Historical Observatory reference

The old #6050 build was also measured five times with the same 500-profile fixture:

| Metric | Observatory median |
|---|---:|
| Network complete | 9.42 s |
| Final UI observed | 12.22 s |
| Total app CPU | 11.66 s |
| Peak combined PSS | 275.1 MiB |
| Peak summed RSS | 503.5 MiB |

This is an architectural reference, not a controlled source-only comparison: the historical Observatory build uses an older app/native base and naturally performs 500 HEAD requests with no separate TCP precheck, whereas upstream and this PR each perform 1,000 GET requests plus 500 TCP prechecks. The matched conclusion is therefore the upstream-baseline-versus-candidate table above.

## Validation

- Play Store debug build installed and exercised on the emulator.
- Verified normal completion, accurate final summary, notification cancellation, concurrency-setting changes, stale-session rejection, and disposable probe-process exit.
- Verified mixed groups on-device for both RealDelay and TCPing: the ordinary SOCKS row received the only result, policy rows remained blank, and the final UI showed `1/1 servers live`. TCPing produced one fixture socket accept.
- Verified a policy-only group reports `0/0 servers live`, produces zero fixture traffic, starts no short-lived Xray core, and exits the disposable service process cleanly.
- Five accepted 500-profile runs for upstream, this PR, and the historical Observatory build, after one warm-up each; exactly one emulator was connected throughout.
- `git diff --check` passes.
- No benchmark harness or test source is included in this branch.

### DIFF
--- a/V2rayNG/app/src/main/AndroidManifest.xml
+++ b/V2rayNG/app/src/main/AndroidManifest.xml
@@ -259,7 +259,7 @@
             android:name=".service.CoreTestService"
             android:exported="false"
             android:foregroundServiceType="specialUse"
-            android:process=":RunSoLibV2RayDaemon">
+            android:process=":Probe">
             <property
                 android:name="android.app.PROPERTY_SPECIAL_USE_FGS_SUBTYPE"
                 android:value="test" />
@@ -269,7 +269,7 @@
             android:name=".service.SubscriptionUpdateService"
             android:exported="false"
             android:foregroundServiceType="specialUse"
-            android:process=":RunSoLibV2RayDaemon">
+            android:process=":SubscriptionUpdate">
             <property
                 android:name="android.app.PROPERTY_SPECIAL_USE_FGS_SUBTYPE"
                 android:value="test" />
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/dto/RealPingBatchStatus.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/dto/RealPingBatchStatus.kt
@@ -1,0 +1,26 @@
+package com.v2ray.ang.dto
+
+import java.io.Serializable
+
+/** One persisted delay-test result delivered from the probe process to the UI. */
+data class RealPingResult(
+    val testId: String,
+    val guid: String,
+    val delayMillis: Long,
+) : Serializable
+
+/** Completed profile count for one active delay-test batch. */
+data class RealPingProgress(
+    val testId: String,
+    val completed: Int,
+    val total: Int,
+) : Serializable
+
+/** Terminal result for one delay-test batch. */
+data class RealPingSummary(
+    val testId: String,
+    val live: Int,
+    val total: Int,
+    val cancelled: Boolean,
+    val listChanged: Boolean = false,
+) : Serializable
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/dto/RealPingEvent.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/dto/RealPingEvent.kt
@@ -3,12 +3,16 @@ package com.v2ray.ang.dto
 sealed class RealPingEvent {
 
     /** Periodic progress update while the batch is still running. */
-    data class Progress(val text: String) : RealPingEvent()
+    data class Progress(val completed: Int, val total: Int) : RealPingEvent()
 
     /** A single server result is available. */
     data class Result(val guid: String, val delayMillis: Long) : RealPingEvent()
 
-    /** The entire batch has finished or been cancelled. */
-    data class Finish(val status: String) : RealPingEvent()
+    /** Terminal counts for the batch. */
+    data class Finish(
+        val live: Int,
+        val completed: Int,
+        val total: Int,
+    ) : RealPingEvent()
 }
 
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/dto/TestServiceMessage.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/dto/TestServiceMessage.kt
@@ -6,7 +6,8 @@ data class TestServiceMessage(
     val key: Int,
     val testId: String = "",
     val subscriptionId: String = "",
-    val serverGuids: List<String> = emptyList(),
+    val serverGuids: List<String>? = null,
+    val excludedServerGuids: List<String> = emptyList(),
     val onlyTcp: Boolean = false
 ) : Serializable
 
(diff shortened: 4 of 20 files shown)


2dust:
感谢pr
这里有个大问题，如下测试：
1. 新建一个分组，这个分组中只有一个策略组，成员来自另外一个分组（500个）
2. 测试这个组，则会测试另外一个分组全部 500 个成员

2dust:
还有就是代码的问题了，为了让策略组测试节省时间，添加了太多的额外代码，这个是不合算的，让维护变得困难。

eliotcougar:
I will look into it.

> 还有就是代码的问题了，为了让策略组测试节省时间，添加了太多的额外代码，这个是不合算的，让维护变得困难。

You think probing the same servers multiple times (in group/out of group) is not a problem? I thought de-duplication was worth it.

If I can have `GetBalancerPrincipleTarget()` from #6109, then probing policy groups will be more straightforward and will look exactly like probing a single server without expanding it to probing all individual servers inside.

PR #6134, opened 2026-08-24
fix: prevent ZIP path traversal during backup restore

eliotcougar:
## Summary

- require every canonicalized ZIP entry target to remain inside a new or empty restore directory
- reject duplicate destinations, file/directory collisions, unsupported compression methods, and inconsistent entry metadata
- bound compressed archive size, entry count, per-entry expansion, total expansion, and compression ratio, with streaming checks against actual extracted bytes
- perform restore I/O on `Dispatchers.IO` and always remove the temporary extraction directory

## Problem

Backup restore previously constructed each output path by appending `ZipEntry.name` to the destination directory without canonical containment validation. A crafted backup selected locally or downloaded from WebDAV could therefore use traversal segments to overwrite files outside the extraction directory.

Extraction also had no resource limits, allowing an archive with excessive entries or expansion to consume disproportionate CPU time and storage.

## Implementation details

The extractor validates the entire central directory before creating the destination:

- maximum compressed archive size: 128 MiB
- maximum entry count: 32
- maximum expanded size per entry: 128 MiB
- maximum total expanded size: 256 MiB
- maximum compression ratio per non-empty entry: 1000:1

The 32-entry cap leaves substantial headroom over the app's current MMKV backup layout, while the 128 MiB per-entry cap accommodates a profile database dominated by an unusually large server list. The compression-ratio and aggregate limits still provide hard upper bounds. Extraction refuses non-empty destinations and existing output files. If streaming fails after extraction begins, newly written output is removed.

The public restore behavior remains unchanged: invalid backups report failure rather than propagating an extraction exception.

## Validation

- `:app:testPlaystoreDebugUnitTest --tests com.v2ray.ang.util.ZipUtilTest`
  - 9 tests covering successful nested extraction, path traversal, archive size, entry count, per-entry size, total expanded size, compression ratio, duplicate canonical destinations, and file/child collisions
- `:app:testPlaystoreDebugUnitTest`

All checks pass for the Play Store debug variant.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/backup/BackupViewModel.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/backup/BackupViewModel.kt
@@ -185,17 +185,20 @@ class BackupViewModel(application: Application) : BaseViewModel(application) {
         }
     }
 
-    private fun performRestore(cacheDir: File, zipFile: File): Boolean {
-        val backupDir = cacheDir.absolutePath + "/${System.currentTimeMillis()}"
+    private suspend fun performRestore(cacheDir: File, zipFile: File): Boolean =
+        withContext(Dispatchers.IO) {
+            val backupDir = File(cacheDir, "restore_${System.nanoTime()}")
+            try {
+                if (!ZipUtil.unzipToFolder(zipFile, backupDir.absolutePath)) {
+                    return@withContext false
+                }
 
-        if (!ZipUtil.unzipToFolder(zipFile, backupDir)) {
-            return false
+                val count = MMKV.restoreAllFromDirectory(backupDir.absolutePath)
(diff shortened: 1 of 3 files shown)


eliotcougar:
Simplification follow-up (`9dc58359`):

- replaced the parallel validated-entry list and target map with one insertion-ordered `LinkedHashMap<File, ZipEntry>`, preserving the full validation pass, archive order, duplicate detection, and file/child collision checks
- made `ExtractionLimits` a plain class and removed Kotlin-internal `@Throws` annotations that did not affect callers
- consolidated repeated rejection-test setup into one helper while retaining all nine named test cases and assertions

This removes 64 net lines without changing the limits or security behavior. The focused ZIP suite, full Play Store debug JVM suite, and Kotlin compilation all pass.

PR #6135, opened 2026-08-24
Make profile storage mutations failure-atomic

eliotcougar:
## Summary

Make checked profile/subscription writes rollback-safe, reject stale updates, and preserve current membership during reorders. Share storage policy across callers without changing the persisted schema or native dependencies.

### Context: #6084 was only partially fixed

Related to #6084, not a claim to close it. Earlier changes improved integrity recovery (#6085), removed delete-before-save replacement (#6093), and added orphan cleanup (#6094). Multi-store operations could still publish inconsistent payloads, indexes, or selection after a failed write; deletion could leave secret-bearing raw configurations behind.

This PR addresses application-level failure windows. It neither establishes the original random `SUB_IDS`/`SUB` loss mechanism nor provides crash-durable transactions across MMKV files. Rollback is attempted after a reported failure; persistent write failure can also prevent restoration.

## Guarantees and structure

- **Checked mutations:** under the `MAIN` multi-process lock, snapshot affected strings and register restoration before mutation. On failure, restore in reverse order and attach rollback failures to the original exception. Prepare profile payloads, subscription metadata, and selection before publishing the authoritative profile index.
- **Deletion:** unpublish IDs before cleaning full-profile, raw-config, and affinity payloads. Interrupted cleanup may leave orphans; an explicitly empty subscription index remains authoritative and does not republish them.
- **Subscription updates:** one pure function validates the expected settings and prepares the value to persist; the manager owns locking and storage access. Settings edits preserve the current `lastUpdated`. Refreshes also require the initiating timestamp to remain unchanged, but may commit an earlier time after a clock correction. Deleted or unpublished subscriptions are rejected.
- **Reorders and migrations:** reorder only current members, retaining additions missing from stale snapshots. Discover legacy subscriptions only when `SUB_IDS` is absent. Merge legacy profile indexes with existing membership; unreadable records and failed required writes leave migrations pending for retry.
- **Callers and caches:** failed saves retain the editor and show existing localized feedback. Serialize drag writes and reconcile visible state after the final queued result. Share cache locking/loading, reuse cached empty groups, and invalidate related group/All caches on refresh.

The standalone transaction helper owns rollback mechanics, not MMKV locking or publication order. No runtime dependency, Android API, or serialized field changes; `kotlinx-coroutines-test` is test-only. Local agent customizations and temporary QA hooks/artifacts are excluded.

## Suggested reading order

1. `StorageMutationTransaction.kt` and its test: checked mutation and rollback mechanics.
2. `SubscriptionUpdateGuard.kt` and its test: eligibility and timestamp ownership.
3. `MmkvManager.kt` and `MmkvPersistenceTest.kt`: locked publication, deletion, and integration regressions.
4. `SettingsManager.kt` / `AngConfigManager.kt`, then repository, ViewModel, and editor changes: migration retries and failure propagation.

## Current validation — `3c0641218`

- Focused subscription-policy, transaction, and manager persistence tests passed.
- `:app:testPlaystoreDebugUnitTest`: 114 tests, no failures/errors/skips.
- `:app:compilePlaystoreDebugKotlin`, `:app:assemblePlaystoreDebug`, and `git diff --check`: passed.
- Regression coverage includes rollback, stale reorders/refreshes, metadata ownership, backward clock changes, authoritative empty indexes, migration retries, queued UI reconciliation, and group caches.
- The MMKV JVM fixture replaces JNI, while executing production transaction/serialization code; it does not prove native durability.

**Not run:** fresh emulator/fault-injection testing for this policy/helper extraction; full keyboard/D-pad/TalkBack coverage; power-loss, mid-operation process termination, or real MMKV/CRC corruption testing.

<details>
<summary>Earlier validation and upstream synchronization</summary>

`744965e0f` merged upstream `2020807c2` (2.3.7), retaining prepared row models alongside cache locking and final-write reconciliation. Its 110 JVM tests, Kotlin compilation, and debug assembly passed.

That merged APK passed Pixel 6a / Android 13 checks for empty/populated/All tabs, profile reorder persistence across restart, and restored order consistency, without an app ANR/crash in captured logs. The previous APK and user state were restored and the emulator stopped. An earlier `40e075d56` cold-boot run had one unexplained Compose-initialization ANR; subsequent launches and the merged run did not repeat it.

Earlier real-MMKV instrumentation exercised payload/index/selection rollback, deletion cleanup and rollback, and delayed refresh after subscription deletion. Temporary fault-injection sources/hooks were removed before publication. These are historical results, not a fresh native-failure run of the current commit.

</details>

### DIFF
--- a/V2rayNG/app/build.gradle.kts
+++ b/V2rayNG/app/build.gradle.kts
@@ -202,6 +202,7 @@ dependencies {
 
     // Testing Libraries
     testImplementation(libs.junit)
+    testImplementation(libs.kotlinx.coroutines.test)
     androidTestImplementation(libs.androidx.junit)
     androidTestImplementation(libs.androidx.espresso.core)
     testImplementation(libs.org.mockito.mockito.inline)
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/handler/AngConfigManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/handler/AngConfigManager.kt
@@ -240,7 +240,12 @@ object AngConfigManager {
      * @param append Whether to append the configurations.
      * @return The number of configurations parsed.
      */
-    private fun parseBatchConfig(servers: String?, subid: String, append: Boolean): Int {
+    private fun parseBatchConfig(
+        servers: String?,
+        subid: String,
+        append: Boolean,
+        subscriptionUpdate: SubscriptionUpdateCommit? = null,
+    ): Int {
         try {
             if (servers == null) {
                 return 0
@@ -264,6 +269,7 @@ object AngConfigManager {
                     configs = configs.map(::ParsedProfile),
                     subid = subid,
                     append = append,
+                    subscriptionUpdate = subscriptionUpdate,
                 )
             }
 
@@ -287,6 +293,7 @@ object AngConfigManager {
         configs: List<ParsedProfile>,
         subid: String,
         append: Boolean,
+        subscriptionUpdate: SubscriptionUpdateCommit? = null,
     ) {
         val keyToProfile = linkedMapOf<String, ProfileItem>()
         val rawConfigs = mutableMapOf<String, String>()
@@ -302,6 +309,7 @@ object AngConfigManager {
             rawConfigs = rawConfigs,
             subscriptionId = subid,
             append = append,
+            subscriptionUpdate = subscriptionUpdate,
         )
     }
 
@@ -313,7 +321,12 @@ object AngConfigManager {
      * @param append Whether to append the configurations.
      * @return The number of configurations parsed.
      */
-    private fun parseCustomConfigServer(server: String?, subid: String, append: Boolean): Int {
+    private fun parseCustomConfigServer(
+        server: String?,
+        subid: String,
+        append: Boolean,
+        subscriptionUpdate: SubscriptionUpdateCommit? = null,
+    ): Int {
         if (server == null) {
             return 0
         }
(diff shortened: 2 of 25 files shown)


eliotcougar:
I've experienced the original issue on the build that included the hardening PR - it wasn't enough... Hopefully, the data loss issues can finally be addressed...

eliotcougar:
Follow-up simplification summary for the last two commits:

- `a5b750a2` simplifies the storage transaction layer itself:
  - removes the extra `prepareThenPublish`/`commit` ceremony while retaining the payload-before-index publication order;
  - replaces the `StoredString` wrapper with nullable snapshots and centralizes string restoration;
  - adds small typed helpers for profile payloads, raw payloads, indexes, selection, and subscription records;
  - unifies single-profile, group, and bulk removal through one locked transaction path;
  - consolidates duplicated index decoding and test mutation setup.
- `a3cf8e79` simplifies callers and UI state handling:
  - shares one `parseAndCommitProfiles` path between batch imports and subscription updates;
  - replaces two loosely related removal booleans with explicit result states;
  - makes the subscription `StateFlow` the sole in-memory list instead of mirroring it in a second mutable collection;
  - removes redundant cache-clearing jobs and waits for the single authoritative refresh;
  - avoids replaying stale server-order snapshots after asynchronous persistence;
  - changes ignored Boolean save/delete callbacks to `Unit` and uses direct function references;
  - shortens the subscription guard and its tests without reducing the covered cases.

Together these commits remove 296 net lines (309 additions, 605 deletions). They are structural simplifications: failure-atomic rollback, fail-closed reads, reference-safe deletion, and authoritative-index publication remain intact.

PR #6137, opened 2026-08-24
Upgrade actions/setup-java to v6

brunoborges:
Upgrade the active build workflow from `actions/setup-java@v5.7.0` to `actions/setup-java@v6`.

Tests were not run (workflow-only change).


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #6161, opened 2026-08-29
TalkBack, 5 of 10: Improve Settings item accessibility

eliotcougar:
## Summary

Part 5 of the TalkBack series improves the existing shared Settings row primitives. Edit, list, menu, and switch rows expose one logical accessibility control, including when disabled. Persistence, screen-specific actions, and visible preference text remain unchanged.

## Accessibility model

- The existing `clickable` and `toggleable` modifiers merge the title and supporting text into the row that owns the action.
- Switch rows use `toggleable(value = checked, enabled = enabled, role = Role.Switch)`. The row owns the observable checked state and the single activation callback.
- The visible `Switch(onCheckedChange = null)` is decoration, not a second action target. Touching the row or its trailing switch, accessibility activation, Enter, Space, and D-pad center use the row action.
- An app-localized On/Off `stateDescription` changes with the checked state, providing native state-change feedback without a custom announcement or moving accessibility focus.
- Disabled rows keep the same semantic structure with `enabled = false`; their label and supporting text stay grouped and the action cannot execute.
- Collapsible groups keep their native Button role. Their localized label includes the group name followed by the expanded/collapsed wording, for example “Core settings. Group is expanded”. The child title's semantics are cleared because the parent label already contains it. This is a combined group label, not a separate `stateDescription`.

The language chooser retains its existing native selectable radio rows. This PR does not change its selection behavior.

No direct accessibility announcements, live-region layer, focus manipulation, custom action menus, or native role suppression are introduced here. Android/TalkBack still controls role and disabled-state wording and the final spoken order.

Relevant official guidance: [Compose API defaults](https://developer.android.com/develop/ui/compose/accessibility/api-defaults), [semantics](https://developer.android.com/develop/ui/compose/accessibility/semantics), and [merging and clearing semantics](https://developer.android.com/develop/ui/compose/accessibility/merging-clearing).

## Known announcement-order limitation

TalkBack may read a switch row's long supporting description before its on/off state. That delays discovery of the setting's most important current information.

Traversal ordering controls the order of separate nodes, not the order of text, state, and role within a merged node. This PR does not put the description into `stateDescription` or replace the switch's native semantics with a forced spoken sentence.

A possible separately reviewed follow-up is to keep the primary switch row concise and move genuinely long explanations behind a named Info action opening a readable modal. Short descriptions could remain inline. **That UI change is not implemented here.**

## Localization and scope

The four resources—group expanded/collapsed labels and On/Off state descriptions—are present in all nine maintained catalogs: English, Arabic, Bengali, Bakhtiari, Persian, Russian, Vietnamese, Simplified Chinese, and Traditional Chinese. App-provided labels use the app locale; native role and disabled wording are supplied by Android/TalkBack.

Production changes are confined to `SettingsItem.kt` and these resource entries. All existing callers benefit without screen-specific callbacks, business-logic changes, or additional state holders. Notification/live-region work and the other focused list PRs remain independent.

## Review refinements

Commit `39637fafc` adds the following refinements:

- Restore the Material minimum-size layout padding lost when the enabled child Switch became non-interactive. This preserves the former switch footprint; the row remains the only toggle. Disabled switches retain their existing footprint.
- Mark collapsible group buttons as headings without duplicating their name or removing the native Button action.
- Replace the version-specific feedback comment with a concise explanation of the observable state description.
- Remove the obsolete three-file Android-only probe harness. Temporary emulator instrumentation stays outside the repository; the resulting PR diff is one production file and nine string catalogs.

## Validation at `39637fafc`

- `:app:testPlaystoreDebugUnitTest`: **57 tests passed**, zero failures, errors, or skips.
- All nine catalogs passed key-presence, uniqueness, and group-name placeholder checks; resource compilation and `git diff --check` passed.
- Android API 37.1 x86_64 emulator with TalkBack enabled: the real Settings switch row exposed one checkable node and one activation target. Native accessibility click, touch on the row and trailing switch, Enter, Space, and D-pad center each toggled and restored its state.
- Native state-description change events were observed for On and Off; accessibility-click activation retained focus on the row.
- Collapsible groups retained one heading and native Button semantics through an expand/collapse round trip.
- The short “Check pre-release versions” row measured **80 dp** and retained one toggle target with immediate native state-change events.
- Disabled “VPN DNS” and “Add HTTP proxy to VPN” rows exposed their disabled state and rejected accessibility activation.
- Pairwise `git merge-tree` checks against all nine companion branches passed in both orders and produced identical trees. This commit also merges cleanly with fetched `upstream/master`.

**Not run:** audio-level verification of the exact spoken phrase/order, a physical-device repeat, older Android versions, and the full light/dark, dynamic-color, single/two-column shared-consumer rendering matrix. The emulator evidence above verifies native nodes, events, focus, and activation; it is not a claim about every TalkBack voice or device.


aliRahimi1997:
The switch state feedback issue is still present in #6161.

I tested the updated switch implementation in #6164 (Per-App Proxy), and the two switches there now behave correctly: after turning a switch on or off, TalkBack immediately reports the new state without requiring the user to leave the control and return to it.

Could #6161 use the same approach/pattern as the updated #6164 implementation for its Settings switches?

There is also another affected switch in the "Check for update" section:
"Off. Check pre-release versions. Switch"

When I toggle this switch, TalkBack does not announce the new state at all. I have to move away from the switch and come back to it to hear whether it is on or off.

So the desired behavior is the same as in the updated #6164 implementation: after toggling a switch, TalkBack should immediately report its new on/off state, while keeping the native switch semantics and without adding custom announcements.

aliRahimi1997:
I retested #6161 from a clean branch based on upstream/master, including the latest commit bdfe08bf.

I can confirm that the Settings switches now provide immediate On/Off feedback with TalkBack, without requiring focus to leave and return to the control.

I also retested the "Check pre-release versions" switch, which now behaves correctly as well.

This matches the improved switch behavior I observed in #6164. My earlier report about the missing immediate state feedback was based on my previous test result and can be considered resolved.

Issue #6153, opened 2026-08-29
2.3.6版本无法区分分组了


HaloKala:
此版本中，订阅分组没有横线区分了，希望回退

2dust:
@eliotcougar  帮忙处理

eliotcougar:
@2dust @xb18 Scrollable tab row was replaced by LazyRow because it was found that it is unusable in accessibility mode. The items that are outside the viewport are skipped by TalkBack navigation. If you want tab dividers, we should add tab dividers to LazyRow items, not simply revert to the old tab row.

I'll make a PR soon...

PR #6163, opened 2026-08-30
Fix group separators without regressing TalkBack navigation

eliotcougar:
## Why

Issue #6153 reported that the Compose group selector no longer provided enough visual separation between subscription groups. Commit 3dc0173b restored the divider and selected indicator by switching the selector back to ScrollableTabRow.

That solves the visual regression, but it also restores the accessibility problem addressed by #6105: TalkBack could not navigate to groups outside the visible tab-row viewport. The keyed LazyRow was introduced because its lazy collection exposes the scrolling semantics needed to reach off-screen groups.

The visual distinction and accessible navigation are independent concerns. The separator should therefore be restored without replacing the lazy collection.

## What changed

- Keep the keyed LazyRow and its off-screen selected-item scrolling.
- Draw a full-width divider beneath the group selector.
- Draw the selected group's 3 dp indicator in the item draw pass so it remains visible above the divider.
- Retain the existing Tab targets, selected state, profile counts, and stable group IDs.

## Validation

- `:app:testPlaystoreDebugUnitTest`
- `:app:assemblePlaystoreDebug -PABI_FILTERS=x86_64`
- Installed the matching-signer APK on a Pixel 6a API 33 emulator.
- Added enough temporary groups to force horizontal overflow; the accessibility UI tree exposed the group selector as scrollable, scrolling revealed off-screen groups, and activating `QA_Group_04` marked it selected.
- Confirmed the full-width divider and selected indicator visually and confirmed an empty crash buffer.
- Enabled and bound TalkBack and verified accessibility focus. ADB-generated touch gestures bypass TalkBack, so they were not counted as full TalkBack gesture-traversal proof.

Follow-up to #6153.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainGroupTab.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainGroupTab.kt
@@ -1,24 +1,33 @@
 package com.v2ray.ang.ui.main
 
 import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.widthIn
+import androidx.compose.foundation.lazy.LazyRow
+import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.ScrollableTabRow
 import androidx.compose.material3.Tab
-import androidx.compose.material3.TabRowDefaults
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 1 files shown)


2dust:
请注意，我们的 app 不是为了给有视觉障碍的人设计的。
如果用很小的代价能兼顾是可以接受的，但是如果为了给某些特定用途而引入大量的维护困难的代码是不可接受的。

eliotcougar:
> 请注意，我们的 app 不是为了给有视觉障碍的人设计的。 如果用很小的代价能兼顾是可以接受的，但是如果为了给某些特定用途而引入大量的维护困难的代码是不可接受的。

TalkBack incompatibility is not the only issue with the old tab-row code. LazyRow also makes scroll-to tab-ID simpler when using "locate profile" (the roll-back probably broke the locate profile, but I didn't check), it is easier to style, and it is in agreement with the official Android UI guidance. 

I'm refactoring my old TalkBack code at the moment. It should be a lot less complicated and mostly limited to ensuring the UI elements have proper semantic properties and associated localized resources.

eliotcougar:
One additional thing I noticed: adding semantics to UI makes automated app testing a lot easier. Robots can use the semantic tree to direct their actions more precisely than when they have to rely on screenshots and tap coordinates.

PR #6165, opened 2026-08-30
TalkBack, 7 of 10: Name profile deletion targets

eliotcougar:
## Summary

Name the specific deletion target in the existing confirmation dialogs for server
profiles, policy groups, proxy-chain profiles, and individual proxy-chain members.
The localized question comes first, followed by the target name on a separate line.

This covers direct and overflow-menu deletion from the main server list and the
registered profile editors. Bulk-deletion questions stay generic.

## Target identity and restoration

- Main-list deletion captures the profile GUID and displayed name together. The
  dialog uses the name only as text; confirmation still removes by GUID. Duplicate
  names and list reordering cannot retarget the pending operation.
- Only these two strings are saved for the pending main-list dialog, using
  `rememberSaveable` and a small `listSaver`. No full profile or action callback is
  serialized.
- Profile-editor dialogs name the saved profile being deleted, not the unsaved
  remarks field. Renaming or clearing a draft therefore does not change the
  identity in the destructive confirmation.
- Pending single-profile confirmations survive activity recreation consistently
  in the main list and the ordinary, custom-config, policy-group, and proxy-chain
  editors.
- A pending proxy-chain member deletion is tied to its stable draft row key. The
  current position and member name are resolved from that key, and confirmation
  removes the member and its key together. A member that has already disappeared
  is not replaced by another member at its former index.

## Scope and accessibility approach

The existing Material confirmation dialog exposes its visible text through native
semantics. There are no custom announcements, focus manipulation, role overrides,
or role suppression.

The small server-menu dispatcher and stable-key member-removal helper are shared
with Part 9 so its custom actions use the same operation paths. This part does not
change server-row semantics, service behavior, deletion policy, or storage format.
Subscription, asset-file, and routing confirmations remain in their own parts.

The unused legacy `ServerActivity` is left unchanged from upstream. Existing
generic confirmation resources remain available to their existing callers.

## Localization

The three named confirmation templates exist in all nine supported catalogs:
English, Arabic, Bengali, Bakhtiari, Persian, Russian, Vietnamese, Simplified
Chinese, and Traditional Chinese. All 27 entries parse as XML and contain exactly
one `%1$s` target placeholder. The review fixes reuse those resources without
introducing new translated text.

## Validation

Reviewed revision: `b8ce732ae6ac56e54169451cbd7c2bc08f29c5f0`.

- `:app:testPlaystoreDebugUnitTest`: 67 tests passed, no failures, errors, or skips.
  This includes 10 focused dispatcher, pending-target saver, and stable-key
  proxy-chain member tests. Saver coverage includes null, duplicate display names
  under different GUIDs, empty names, non-Latin text, and embedded newlines.
- `:app:compilePlaystoreDebugKotlin`: passed.
- `:app:assemblePlaystoreDebug -PABI_FILTERS=x86_64`: passed.
- Pixel 9 Pro emulator, API 37.1, with TalkBack enabled: inspected native dialog
  text and exercised native accessibility click actions in ordinary, custom,
  policy-group, and proxy-chain editors. Unsaved renames and blank remarks did not
  change the saved deletion target. Cancel preserved each profile; actual
  activity recreation preserved each pending dialog. These emulator checks used
  the identical source tree immediately before committing; source-tree equality
  was checked before the commit.
- Main-list emulator check: recreated a pending dialog, confirmed it, and checked
  through the existing storage owner that only the captured GUID was removed.
  Another profile and the selected-server GUID remained unchanged.
- Used a temporary external QA harness, not an `androidTest` source directory in
  the PR. Removed its test fixtures and restored the emulator's original APK and
  accessibility settings. The existing lab subscription was retained.
- Pairwise `git merge-tree` checks against fetched upstream and all nine companion
  branches passed in both orders and produced identical trees.
- `git diff --check`: passed. No instruction files or generated inputs are in the
  feature diff.

### Not run

- Recorded TalkBack speech, swipe focus order, physical touch, keyboard, and D-pad
  interaction. Native-tree inspection and accessibility action execution do not
  establish exact spoken output or gesture behavior.
- Other Android versions, physical devices, release builds, other ABIs, and the
  full locale/theme/layout matrix. The emulator run used one English x86_64 debug
  configuration; localized resources were checked structurally, not spoken in
  every language.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainDialogs.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/ui/main/MainDialogs.kt
@@ -5,6 +5,11 @@ import androidx.compose.ui.res.stringResource
 import com.v2ray.ang.R
 import com.v2ray.ang.ui.compose.DeleteConfirmDialog
 
+data class ServerDeleteTarget(
+    val guid: String,
+    val profileName: String,
+)
+
 @Composable
 fun MainDialogs(
     showDelAllConfirm: Boolean,
@@ -16,7 +21,7 @@ fun MainDialogs(
     showDelInvalidConfirm: Boolean,
     onDismissDelInvalid: () -> Unit,
     onConfirmDelInvalid: () -> Unit,
-    showRemoveConfirm: String?,
+    showRemoveConfirm: ServerDeleteTarget?,
     onDismissRemove: () -> Unit,
     onConfirmRemove: (String) -> Unit,
 ) {
@@ -41,11 +46,10 @@ fun MainDialogs(
             onDismiss = onDismissDelInvalid
         )
     }
-    if (showRemoveConfirm != null) {
-        val guid = showRemoveConfirm
+    showRemoveConfirm?.let { target ->
         DeleteConfirmDialog(
-            message = stringResource(R.string.confirm_delete_profile),
-            onConfirm = { onConfirmRemove(guid) },
+            message = stringResource(R.string.confirm_delete_profile_named, target.profileName),
+            onConfirm = { onConfirmRemove(target.guid) },
             onDismiss = onDismissRemove
         )
     }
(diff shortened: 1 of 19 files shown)


eliotcougar:
Almost everything has been addressed in the previously extracted PR. Only one fix is left here.

PR #6166, opened 2026-08-30
TalkBack, 8 of 10: Improve shared navigation and action labels

eliotcougar:
## Summary

Improve shared navigation and action labels without replacing native controls:

- Read the visible page title before Back or Menu within the shared top app bar.
- Hide the decorative drawer header and redundant app-name text from accessibility.
- Name the Menu button's native click action **Open navigation menu**.
- Name the service button's native click action **Connect** or **Disconnect**, while
  retaining its existing **Start service** / **Stop service** accessible name.
- Give App Picker's overflow button the existing localized **More options** name.
- Localize **Clear search** and make Logcat's copy action say **Copy log to clipboard**.

## Implementation

The title has `traversalIndex = -1f`. Material 3 already makes `TopAppBar` a traversal
group, so there is no duplicate group modifier. This changes order only within the
bar; it does not request accessibility focus or rewrite window announcements.
Search mode keeps its existing text-input behavior. The small title-index modifier
keeps the shared top-bar imports separate from Part 6's `AppListItem` changes.

Both the drawer header container and its text use `hideFromAccessibility`. These
are separate semantic nodes; hiding the container does not by itself hide its
child text in the pinned Compose version. Drawer destinations keep Material's
native roles, labels, selection semantics, and activation. Native drawer dismissal
is unchanged.

Menu and service-button action labels augment `onClick` semantics with a label and
a null action. Their existing `IconButton` / `FloatingActionButton` callbacks still
own activation. This does not introduce another click handler, duplicate custom
action, role suppression, or manual TalkBack announcement. Accessibility services
remain responsible for presenting their action menu and gesture hints.

Official guidance:
[traversal ordering](https://developer.android.com/develop/ui/compose/accessibility/traversal)
and [hiding decorative semantics](https://developer.android.com/develop/ui/compose/accessibility/merging-clearing).

## Scope and compatibility

No service lifecycle, repository, subscription-update, paging, or server-selection
behavior changes. Main server rows, group-tab names/counts, and delay units remain
in Part 9; connection-test status and live-region feedback remain in Part 10.
This part does not touch `MainGroupTab`, `MainServerPager`, or `MainViewModel`.

The existing bottom-bar import and label placement is deliberately retained to
avoid overlap with Part 10. The redundant traversal-group helper and this PR's two
permanent Android-test files have been removed. UI checks are maintained as local
external QA, not test scaffolding in the contribution.

## Localization

Clear search, Copy log to clipboard, Connect, and Disconnect are supplied in all
nine supported catalogs: English, Arabic, Bengali, Bakhtiari, Persian, Russian,
Vietnamese, Simplified Chinese, and Traditional Chinese. Menu and More options
reuse existing translated resources. No new translation changes were needed for
the review refinements.

## Validation

Reviewed revision: `5ab00cd5642518c8382ccc3f191e6f6bf5c552db`.

- `:app:testPlaystoreDebugUnitTest`: 57 tests passed, no failures, errors, or skips.
- `:app:compilePlaystoreDebugKotlin` and
  `:app:assemblePlaystoreDebug -PABI_FILTERS=x86_64`: passed. Final Gradle checks
  were repeated on the committed revision.
- Pixel 9 Pro emulator, Android API 37.1, with TalkBack enabled: checked native
  title-before-navigation traversal links on Config, Settings, Subscriptions,
  Logcat, About, and App Picker. The title ordering remained intact without the
  redundant traversal-group modifier.
- Confirmed the Menu button's native action label, button role, and single click
  target; opened the drawer through accessibility click, injected touch, Enter,
  Space, and D-pad center. Confirmed native dismissal and that no visible branding
  or silent header precedes the drawer destinations.
- Checked localized search-clear naming, clearing the query, returning to the
  normal title, and App Picker's accessible overflow name.
- Parsed all nine catalogs and checked all 36 changed action entries plus the
  nine reused Menu-action entries for presence, uniqueness, and nonempty text.
- Pairwise merge simulations against fetched upstream and each of the other nine
  accessibility branches passed in both orders with identical trees. These are
  textual compatibility checks, not a full combined-build test.
- `git diff --check`: passed. No Android-test scaffolding, generated native inputs,
  or instruction changes remain in the PR diff.

### Not run

- Recorded TalkBack speech and physical TalkBack swipe/double-tap gestures. Native
  traversal links, node properties, and injected input do not prove exact spoken
  wording or actual TalkBack gesture behavior.
- The full Android-version, physical-device, locale, RTL, dark-theme,
  dynamic-color, and single/double-column layout matrix. Runtime checks used one
  English x86_64 emulator configuration.
- A new combined APK build or service start/stop lifecycle run. This refinement
  leaves the bottom-bar service labels and implementation unchanged.


aliRahimi1997:
I tested PR #6166 on a real device with TalkBack enabled.

Most of the shared navigation and action labels work correctly. I found two issues in the navigation drawer:

1. Initial accessibility focus

When the navigation drawer is opened, TalkBack initially places accessibility focus on a node that speaks nothing. I have to swipe once before the first actual navigation item is announced.

Ideally, when the drawer opens, accessibility focus should start on the first meaningful navigation item rather than on a silent/non-actionable node, if this can be achieved using the standard Android/Compose accessibility mechanisms.

2. Navigation item role

The navigation items are announced with "Tab" at the end. For example:

"Settings, tab"
"Per-app proxy, tab"

The items otherwise work correctly.

Could you please check whether this "Tab" role is actually the intended semantic role for these NavigationDrawerItem entries, or whether it is being introduced by the accessibility changes in this PR?

I am intentionally not suggesting a custom TalkBack announcement or manually manipulating accessibility focus. I would prefer to preserve the native semantic contract and fix the underlying semantics/structure if necessary.

Other shared action labels in this PR worked correctly in my testing.

eliotcougar:
I will address the first issue. However, the second one has to stay. Tab is the default role of the Navigation drawer items. I intentionally did not override the roles anywhere in the new implementation.

PR #6168, opened 2026-08-30
TalkBack, 10 of 10: Announce connection test feedback

eliotcougar:
## Summary

Part 10 separates connection-test feedback from the continuously changing visual status:

- The bottom row exposes Connected or Not connected and a native, localized Check connectivity action only while connected.
- Current-test and bulk-test progress remain visible but inert to TalkBack.
- An assertive live region publishes Testing, then the final current-test result or bulk completion. Fast results cannot replace the start message before it is published.
- After the current result's live-region message retires, that result becomes available on the status row for subsequent navigation. Success/failure, full localized delay units, and available country/IP information are preserved.

## Live-region implementation

Uses the exact same `AccessibilityLiveRegionText` implementation as Part 1, including locale spans, stable hidden idle node, frame-separated publication, and publication acknowledgement. The identical shared file keeps both branches independently buildable and avoids an add/add conflict.

Test messages are queued in publication order and held for one second after publication acknowledgement. This is a semantics lifetime, not a claim that the application can detect when TalkBack has finished speaking. Collection is limited to the resumed screen. Leaving it clears transient messages; a result received in the background remains durable ViewModel state and can be read upon return without replaying old speech.

There are no direct accessibility announcements, focus requests, native-role suppressions, or new interactive Snackbar/Toast nodes. Repeated identical tests retain distinct event IDs.

## Request ownership and cancellation

- Current and bulk requests carry independent UUIDs. Replies from older requests cannot finish newer tests, including repeated bulk tests in the same group.
- A bulk test cancelled from its notification reports cancellation rather than successful completion. Empty service-side batches also retire without false completion feedback.
- Current tests use daemon acknowledgement, so a missing daemon does not leave the UI indefinitely testing.
- Native current probes have an owned scope. Replacing a probe, stopping the core, or reloading it cancels pending feedback. Cancellation checks prevent an old native result from being published, and a request-specific cancellation reply retires the UI's pending state. Cancellation does not claim to interrupt a blocking JNI call immediately.
- Request IDs are optional Intent metadata. Existing serialized result and test-command payloads are unchanged; native library APIs and persisted data formats are unchanged.
- State transitions and publication IDs stay on the main thread; cache reloads and native/network work remain on background dispatchers.

## Localization and scope

All nine supported catalogs contain the completion, stable connected-state, and spoken delay resources. Russian connection-delay phrasing uses the accusative singular, independently of the server-list delay resource. No XML catalogs were split, and this revision changes no translation text.

The permanent Android test fixture and copy-only status helpers were removed. Focused JVM regressions cover queue ordering, repeated results, cancellation, stale publication callbacks, and current/bulk request ownership. Temporary Android instrumentation is outside the PR.

## Validation

Validated revision: `94fe07227` (same tree as the merge-test candidate).

- 64 JVM tests passed, including the seven focused feedback/request regressions.
- Play Store debug Kotlin compilation and x86_64 APK assembly passed.
- An external Android 13 / TalkBack-enabled probe exercised the real screen and ViewModel with controlled current-test replies. It verified same-main-thread-turn start/result ordering, repeated success and failure, delayed result exposure, stale replies after stop/replacement, and background-result retention without replay.
- Real repository broadcasts verified matching bulk request IDs, rejection of stale completion, inert progress, and terminal completion publication.
- The actual CoreTestService notification-cancel command path cleared testing without announcing Testing complete.
- The real ordered-broadcast daemon-absent path retired a pending test without fabricating a result.
- All three added resource keys exist exactly once in all nine locale catalogs.
- Twenty pairwise merge simulations passed against current upstream and Parts 1–9 in both orders. The Part 1 shared live-region file is byte-identical after Git normalization; the merged repository and ViewModel event paths were inspected.
- Whitespace checks passed. Generated native inputs, APKs, temporary instrumentation, and local instructions are not included.

## Not run

- Physical-device speech/voice and full spoken-utterance timing checks. The emulator probe preserved TalkBack and inspected native accessibility content/events; it does not assert that every phrase was audibly completed.
- Actual blocking JNI cancellation and stop/reload races across VPN, proxy-only, and root modes. Request invalidation was checked through controlled replies; the native cancellation path was compiled and reviewed.
- Full theme, layout, keyboard/D-pad, Android-version, and locale-fluency matrices; no visual layout or activation handler was redesigned.

The emulator's original APK and accessibility settings were preserved, with existing subscriptions and profiles left intact.


aliRahimi1997:
I tested this PR with TalkBack.

The connection-test result itself is announced correctly when the test finishes, and there is no problem with that announcement.

The issue is that the result does not remain available on the “Check status” button after the test completes.

For example, if the test reports “Connected” and I hear it at that moment, everything is fine. However, if I do not hear the announcement, there is no result left on the button for TalkBack to read later. I have to run the connection test again to get the result announced again.

So the problem is not that the button cannot read its status when it has one. The problem is that the test result is not retained as persistent accessible information on the button after the test finishes.

Could you please check whether the latest connection-test result can remain exposed through the appropriate accessibility semantics/content of the button after the test completes, while keeping the existing one-time announcement?

I did not observe any other issue with the connection-test announcement.

eliotcougar:
The testing progress and result was hidden from TalkBack to prevent duplicate announcements (from the status bar and the live region at the same time). I guess, we can reset the hidden state after it was spoken once.

PR #6171, opened 2026-08-31
Fix routing-rule creation and crash caused by duplicate IDs

eliotcougar:
## Summary

Fix two routing-rule identity issues in a focused branch based on current `master` (`c548e5f615d87a9220fd066b9f31fa127551ff90`, **Use ruleset IDs for routing edits**).

### 1. Adding a routing rule immediately closes the editor

The Add action intentionally starts `RoutingEditActivity` without `ruleset_id`. After the ID-based editing change, the activity resolved that to `initial == null` and unconditionally called `finish()`. Thus a valid new-rule request was treated as a stale edit request.

The editor now distinguishes these cases:

- **No ID:** open a blank editor and create a rule on Save.
- **Existing ID:** load and edit that rule, regardless of its current list position.
- **Explicit ID that no longer exists, including an empty ID:** close without creating or modifying another rule.

### 2. Imported or previously saved duplicate IDs can crash the routing list

The routing list and its reorderable rows use `RulesetItem.id` as a Compose key. Importing an exported ruleset retains the current locked rules and previously appended the entire import, including copies of those same locked rules with the same IDs. The old reload code repaired empty IDs only. Nonempty duplicates therefore reached Compose and also made ID-based edit/delete ambiguous.

This duplicate-ID exposure predates the latest editing change; it is **not a repeated-title problem**. Rules with identical remarks must remain supported.

## Fix

- Enforce nonblank, unique IDs when reading and writing routing lists through the existing `MmkvManager` storage boundary.
- Repair legacy missing/null/blank IDs and duplicate IDs, preserving the first valid occurrence, all rule contents, and list order. Persist repaired identities before exposing the list to callers; do not publish ephemeral replacements when that write fails.
- Keep retained locked rules without re-appending their exact exported copies. Preserve different imported rules even when their titles or IDs collide; assign fresh IDs rather than discard those rules.
- Use a non-string header key so an imported ID such as `domain_strategy` cannot collide with non-rule content.
- Remove the per-row ID migration from the list ViewModel. Load off the UI thread and cancel obsolete reloads, including reloads superseded by a toggle/reorder.
- Give the editor its own lifecycle-owned ViewModel, consistent with the upstream UI guide: loaded state is a lifecycle-collected `StateFlow`, storage runs off the UI thread, and the form builds an edited copy rather than reading MMKV during composition. Uncommitted text remains `rememberSaveable` form state.

No routing/export fields are renamed or removed. No rules are deduplicated by title. No native dependency, Shizuku, TV, Logcat, or combined-branch changes are included.

## Validation

- **18 focused JVM regression tests passed**, including legacy JSON, null/blank/duplicate IDs, stable persistence across reload, locked export/import round trips, repeated titles, Add vs stale edit, ID-based edit/delete after reorder, read/write failure handling, and pending reload vs toggle/reorder.
- **Full Play Store debug unit suite: 75 tests passed.**
- `:app:compilePlaystoreDebugKotlin` and `:app:assemblePlaystoreDebug` passed.
- Installed the matching-signer x86_64 Play Store debug APK on the **Pixel 9 Pro API 37.1 emulator**. Verified Add → Save → reopen, saving with the IME visible, two independent rules with the same remarks, locked-rule export/reimport twice, editing only the intended rule, unsaved input surviving activity recreation, keyboard Save, and D-pad Delete/confirmation. No application crash appeared in the crash buffer.
- `git diff --check` passed.

### Existing lint failures

`:app:lintPlaystoreDebug` ran and reports **4 errors, 302 warnings, and 1 hint**. The four errors are API-level checks in unchanged upstream code: `Process.waitFor(timeout, unit)` in `RootManager.kt` and `RootShell.kt`, and `windowLightNavigationBar` in the day/night theme resources. This PR adds no lint suppressions or baseline and does not bundle unrelated fixes for those findings.

**Not run:** runtime validation on physical hardware or other Android versions, TalkBack spoken-output testing, and F-Droid builds. The changes do not branch by distribution; runtime interaction evidence is limited to the emulator above.

### DIFF
--- a/V2rayNG/app/build.gradle.kts
+++ b/V2rayNG/app/build.gradle.kts
@@ -202,6 +202,7 @@ dependencies {
 
     // Testing Libraries
     testImplementation(libs.junit)
+    testImplementation(libs.kotlinx.coroutines.test)
     androidTestImplementation(libs.androidx.junit)
     androidTestImplementation(libs.androidx.espresso.core)
     testImplementation(libs.org.mockito.mockito.inline)
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/handler/MmkvManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/handler/MmkvManager.kt
@@ -30,8 +30,28 @@ import com.v2ray.ang.util.Utils
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.drop
+import java.util.UUID
 
 internal class ProfileStorageException(message: String) : IllegalStateException(message)
+
+/** Keep the first occurrence of an ID; repair only missing or colliding identities, never rule contents. */
+internal fun ensureRoutingRulesetIds(rulesets: List<RulesetItem>): Boolean {
+    val reserved = rulesets.mapTo(mutableSetOf()) { it.id }
+    val seen = mutableSetOf<String>()
+    var changed = false
+    for (ruleset in rulesets) {
+        if (ruleset.id.isNullOrBlank() || !seen.add(ruleset.id)) {
+            var id: String
+            do {
+                id = UUID.randomUUID().toString()
+            } while (!reserved.add(id))
+            ruleset.id = id
+            seen.add(id)
+            changed = true
+        }
+    }
+    return changed
+}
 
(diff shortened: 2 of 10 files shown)


2dust:
https://github.com/2dust/v2rayNG/commit/1af8f4677ccee032957aec63a328db9e96bf9a2f

PR #6172, opened 2026-08-31
Prevent duplicate subscription IDs from crashing Compose lists

eliotcougar:
## Summary

Prevent duplicate entries in the persisted subscription index from producing duplicate Compose keys. The production change is one `distinct()` call in `MmkvManager.decodeSubsList()`, with a short explanatory comment and six regression tests.

## Why

Subscription records are stored by GUID, while their display order is stored separately in the `MAIN/SUB_IDS` JSON array. If that array contains repeated GUIDs, `decodeSubscriptions()` currently emits the same record repeatedly. Both the subscription `LazyColumn` and its reorderable items use that GUID as their key, so a damaged index can cause a duplicate-key Compose crash.

For example, with existing records `a`, `b`, and `c`:

```text
Saved index:       [b, a, b, c, a]
Before:            [b, a, b, c, a]
After:             [b, a, c]
```

This is defensive handling of already-duplicated stored data, not a claim that creating subscriptions with the same name produces duplicate IDs.

## Scope and behavior

- Normalize the shared reader so subscription rows and other index consumers receive unique GUIDs.
- Preserve the first occurrence and existing order, and keep returning a mutable list for current callers.
- Do not change subscription records, names, IDs, Compose keys, or the persisted format.
- Do not write back during reads. Existing normal index writes can subsequently persist the deduplicated list.
- Preserve distinct subscriptions with equal or blank names and the existing empty/malformed-index fallback.

No UI, localization, native-library, or dependency changes. Based directly on current `upstream/master`; independent of the accessibility changes in #6158.

## Validation

- Ran the six new regression tests against unchanged upstream: three failed, exposing repeated decoded IDs/rows and a surviving duplicate after removal.
- With the fix, all six pass: `:app:testPlaystoreDebugUnitTest --tests com.v2ray.ang.handler.SubscriptionIndexTest`.
- Full `:app:testPlaystoreDebugUnitTest`: **63 tests passed**.
- API 37.1 Pixel 9 Pro emulator: injected repeated GUIDs into real MMKV storage; verified exactly three rows in first-occurrence order, correct edit target, and safe list reopening. A subsequent device-side assertion confirmed that the raw saved index still contained duplicates while both readers returned unique IDs. No app crash was observed.
- Removed the temporary fixtures and restored the emulator's pre-test APK and original subscription index.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/handler/MmkvManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/handler/MmkvManager.kt
@@ -718,7 +718,8 @@ object MmkvManager {
         return if (json.isNullOrBlank()) {
             mutableListOf()
         } else {
-            JsonUtil.fromJsonSafe(json, Array<String>::class.java)?.toMutableList() ?: mutableListOf()
+            // Keep the first occurrence so a damaged index cannot produce duplicate Compose keys.
+            JsonUtil.fromJsonSafe(json, Array<String>::class.java)?.distinct()?.toMutableList() ?: mutableListOf()
         }
     }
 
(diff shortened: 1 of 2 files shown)


2dust:
会存在重复的ID 吗？
先合并

PR #6169, opened 2026-08-30
Update Luri Bakhtiari translation

hosseinabaspanah:

PR #6174, opened 2026-08-31
Ask for a group name when importing subscriptions

eliotcougar:
Clipboard and QR imports currently save new subscription groups immediately. Links without a name all receive the same `import sub` placeholder, leaving users with indistinguishable groups to rename afterward. A batch can contain several subscription URLs, so the naming prompt also needs to identify the URL being imported.

This change asks for a name before saving each new subscription group. The dialog has no visible heading and shows **Provide the name for the subscription group: [URL].**, one prefilled text field, and **OK / Cancel** buttons.

- Suggest the URL fragment's name when available, or a localized placeholder otherwise. Choose the first available suffix when the suggestion already exists: `import sub 2`, `import sub 3`, and so on.
- Save only after OK, trim the confirmed name, and disable OK for blank input. Cancel or dismiss skips only the current subscription; a batch continues with the next URL. Previously confirmed groups remain saved.
- Keep the URL and draft name together in the main ViewModel, preserving their association across activity recreation. Key the dialog by URL so each item in a batch gets its own initial focus.
- With TalkBack enabled, let the URL sentence receive initial accessibility focus. Retain a localized accessible pane title without a visible heading, and avoid automatically focusing the editor. Without touch exploration, retain immediate typing in the name field.
- Skip existing subscription URLs, including a recheck after confirmation. When only existing subscriptions are supplied, show localized singular/plural duplicate feedback instead of a generic failure.

Clipboard, QR, and local-file actions share this pipeline. The QR scanner shortcut uses the same confirmation flow while retaining its target-group and append behavior. Ordinary proxy profiles import without prompting, and URL-scheme imports retain their automatic behavior. No existing groups are renamed; storage formats and native dependencies are unchanged. All dialog strings are localized in the nine supported catalogs, with bidirectional formatting around URLs in right-to-left layouts.

Bakhtiari subscription-import wording now matches the [native-speaker correction from hosseinabaspanah in PR #6169](https://github.com/2dust/v2rayNG/pull/6169#issuecomment-5512959269). Four values changed; the field label and singular duplicate message already matched. Resource IDs, plural categories, and the URL placeholder are unchanged.

For the wording-only update at `424b4d32`, XML parsing, exact comparison against all six supplied values, and `git diff --check` passed. **Not run:** builds and tests for this update, as requested.

Earlier functional validation on `7250f15a` (before the Bakhtiari wording correction):

- All 62 Play Store debug JVM tests passed; Kotlin compilation, Play Store debug assembly, and Android-test assembly passed.
- All seven subscription-import instrumentation tests passed on Pixel 6a / Android 13 (API 33), x86_64, with TalkBack enabled and disabled. Coverage includes actual clipboard import, three sequential unnamed URLs, initial accessibility/editor focus, localized message and pane title, OK/Cancel, suffix reuse, recreation, duplicate/concurrent imports, Base64 batches, and ordinary proxy imports.
- The three-URL focus test also passed in Russian and Arabic. It asserts actual native accessibility focus on the URL sentence without requesting focus from the test.
- Checked the heading-free English and Arabic layouts, longer URL wrapping, and the English dialog with the IME visible. Runtime QA used a separate package to preserve the installed app's data. Earlier validation on `477c6b4c` also covered touch confirmation, persisted names after reopening, Tab/Enter cancellation, and D-pad button traversal/activation.
- Earlier validation exercised QR decoding through the scanner's image picker; camera decoding code is unchanged.

Not run: physical-camera QR scanning, end-to-end TalkBack touch gestures or spoken-audio verification, and speech fluency in every locale. The accessibility checks establish native focus, text, title, and action behavior, not complete spoken traversal.

Known accessibility limitation: the editable field's label can be encountered during character/word navigation in the current Compose implementation. Field labeling is unchanged here, pending stable native hint support across the common text fields. No alpha dependency upgrade or editable-field content-description replacement is included. See the [follow-up explanation](https://github.com/2dust/v2rayNG/pull/6174#issuecomment-5508724711).

### DIFF
--- a/V2rayNG/app/src/androidTest/java/com/v2ray/ang/ui/main/SubscriptionImportTest.kt
+++ b/V2rayNG/app/src/androidTest/java/com/v2ray/ang/ui/main/SubscriptionImportTest.kt
@@ -5,7 +5,12 @@ import android.content.ClipData
 import android.content.ClipboardManager
 import android.os.Bundle
 import android.util.Base64
+import android.util.Log
+import android.view.View
+import android.view.accessibility.AccessibilityManager
 import android.view.accessibility.AccessibilityNodeInfo
+import androidx.core.text.BidiFormatter
+import androidx.core.view.accessibility.AccessibilityNodeInfoCompat
 import androidx.lifecycle.ViewModelProvider
 import androidx.test.core.app.ActivityScenario
 import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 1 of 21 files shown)


aliRahimi1997:
Thanks for the changes. I tested the subscription import flow with TalkBack and found two accessibility/usability issues worth mentioning:

1. Accessibility issue with text fields

When TalkBack focus is inside an EditText, using character-by-character or word-by-word text navigation causes the field's label to be announced as part of the text, always appearing after the text entered by the user.

For example, in the new "Subscription name" field, after entering a name, navigating through the text also announces "Subscription name" at the end. I can reproduce the same behavior in other EditTexts in v2rayNG, such as the manual subscription import field, so this does not appear to be specific to this PR. I also tested other Android apps and did not observe the same behavior there.

2. Duplicate subscription error message

When attempting to import a subscription that already exists, the duplicate is correctly prevented. However, TalkBack only announces "Failed", without explaining the reason for the failure.

For a screen reader user, this makes it unclear that the subscription was rejected because it already exists. A more descriptive error message would make the reason for the failure much clearer.

eliotcougar:
Thank you, @aliRahimi1997, for reporting both issues.

### Duplicate subscription feedback: fixed

Commit a83d95812191172bf5ec32bda3de71adcdf84764 fixes item 2. Previously, the import result only retained successful counts, so an existing subscription could fall through to the generic failure message. The result now also carries duplicate counts, including Base64 input and the recheck after the naming dialog. The main import flow displays “This subscription already exists.” or its plural instead of “Failed”, using localized resources in all nine catalogs and the existing feedback mechanism.

Validation: all 62 JVM tests and six instrumentation tests passed on Pixel 6a / Android 13. The device tests retained TalkBack and checked the duplicate-feedback text in the accessibility tree. This is not a claim of verified speech in every locale.

### Text-field label during character/word navigation: deferred

Item 1 remains acknowledged and unresolved. In the inspected accessibility tree, the editable value was only the entered name; Material's “Subscription name” label appeared as a descendant text node. This is consistent with your report and also affects existing fields, rather than being specific to this dialog.

This branch uses Compose UI 1.11.4 / Material3 1.4.0. Compose's native `hintText` semantics, mapped to `AccessibilityNodeInfo.hintText`, were introduced in [Compose UI 1.13.0-alpha01](https://developer.android.com/jetpack/androidx/releases/compose-ui#1.13.0-alpha01). We will defer this work until the project can adopt a stable Compose version containing the relevant support; this PR will not upgrade to alpha dependencies.

We also will not substitute a custom `contentDescription` for the editable field, hide its purpose, strip its editing/selection semantics, or add focus/announcement workarounds. [Google's editable-view guidance](https://support.google.com/accessibility/android/answer/6378120?hl=en-GB) explains that `contentDescription` can interfere with describing and navigating entered text; native hints/label associations are the appropriate direction.

A future Compose update is not being presented as an automatic fix. After adopting stable support, we should review the common field-label implementation and verify empty and populated fields, character/word navigation, cursor movement, and selection with TalkBack. Until then, the standard editable-field behavior is deliberately unchanged.

2dust:
感谢
但是这个功能也是完全没有必要

PR #6185, opened 2026-09-04
Fix adding routing rulesets and harden ruleset id handling

patterniha:
Since ruleset lookups moved from list index to id, RoutingEditActivity finished itself whenever no ruleset was found. The add flow starts the activity without a ruleset_id on purpose, so the "+" button opened and immediately closed the screen. Only finish when an id was given but no longer resolves.

Ids were also not guaranteed to be unique. Presets and imported lists carry none, and exporting to the clipboard and importing the result next to the locked rulesets it was exported with repeats their ids. Duplicate ids make the keyed list on the routing screen throw and send id-based edits and deletes to the wrong ruleset. Assign missing or duplicated ids when rulesets are imported and when the list is loaded, in one write instead of one save per ruleset.

Finish the migration to ids: the enabled toggle saves by id, the unused index-based overload is gone, the id-based functions no longer shadow their parameter, and their formatting matches the rest of the file.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/handler/SettingsManager.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/handler/SettingsManager.kt
@@ -30,6 +30,7 @@ import com.v2ray.ang.util.LogUtil
 import com.v2ray.ang.util.Utils
 import java.io.File
 import java.io.FileOutputStream
+import java.util.UUID
 import kotlin.random.Random
 
 object SettingsManager {
@@ -70,6 +71,7 @@ object SettingsManager {
         }
 
         return JsonUtil.fromJsonSafe(assets, Array<RulesetItem>::class.java)?.toMutableList()
+            ?.also { ensureRoutingRulesetIds(it) }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


eliotcougar:
Review note from **GPT-5.6 Sol**, working with @eliotcougar:

This is a high-quality and impressively compact implementation. In particular, saving the enabled toggle by ID and removing the obsolete index-based overload finish the ID migration more cleanly than #6171 does. The Add-versus-stale-edit distinction is also expressed much more directly. Thank you for putting this together.

I think a few correctness details still need to be incorporated before this fully closes the duplicate-identity/key problem. The corresponding small safeguards can be borrowed from #6171 without taking its broader editor ViewModel and asynchronous-loading refactor:

1. **Preserve every existing valid ID while repairing the list.**

   `ensureRoutingRulesetIds()` currently knows only the IDs already visited. A repaired blank/duplicate entry can therefore be assigned an ID belonging to a valid later entry; the later entry is then treated as the duplicate and loses its original identity. A generated UUID also is not retried if `seenIds.add()` fails. #6171 first reserves all original IDs, then generates in a loop until `reserved.add(id)` succeeds. That preserves every originally valid unique ID and makes the invariant unconditional rather than probabilistic.

2. **Keep rule keys in a different namespace from non-rule LazyColumn keys.**

   A unique, nonblank imported rule ID can legally be `"domain_strategy"`. The repair accepts it, but `RoutingSettingActivity` already uses the same string as the header key, so Compose can still receive duplicate keys and throw. #6171 changes the header to a non-string key (for example `item(key = 0)`), which is a tiny and sufficient correction.

3. **Do not retain an identical exported copy of an already-retained locked rule.**

   `resetRoutingRulesetsCommon()` retains the current locked rules and then appends the entire imported list. When that list was exported by the app, it contains those locked rules again. Reassigning the copied IDs prevents the key crash, but leaves duplicate rule contents and repeats them on subsequent imports. #6171 filters only exact copies of retained locked rules while deliberately preserving different rules with the same title or colliding ID.

4. **Enforce and persist the identity invariant at the storage boundary.**

   The current repair is invoked by selected SettingsManager/import/list-loading paths. Another present or future caller of `MmkvManager.decodeRoutingRulesets()` or `encodeRoutingRulesets()` can still observe or store invalid identities. Also, if the repair write fails, the routing screen publishes temporary in-memory IDs that do not exist in storage; a later ID-based save can then target the wrong data or append another rule. #6171 centralizes repair in decode/encode and does not expose a repaired list unless the migration was persisted successfully.

I would retain the compact structure of this PR and borrow those focused pieces from #6171. The lifecycle-owned editor ViewModel and reload-cancellation work in #6171 are useful architectural hardening, but they are not necessary to fix this bug and need not enlarge this PR.

2dust:
原来的用 id 模式已经回滚了

PR #6195, opened 2026-09-06
i18n: Localize notification-channel names

eliotcougar:
Replace hard-coded channel labels with dedicated resource names for the main service, subscription updates, and connection checks in all nine supported locales. Resolve names through AppLocaleManager for service and application contexts.

Reuse the existing channel when refreshing its localized name, preserving channel IDs and user-controlled importance, sound, vibration, and visibility. Apply initial defaults only to newly created channels and remove the obsolete hard-coded service-channel name.

### DIFF
--- a/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
+++ b/V2rayNG/app/src/main/java/com/v2ray/ang/AppConfig.kt
@@ -200,7 +200,6 @@ object AppConfig {
     /** Notification channel IDs and names. */
     // Use a new ID because Android does not let an app raise an existing channel's importance.
     const val RAY_NG_CHANNEL_ID = "CORE_M_CH_ID_V2"
-    const val RAY_NG_CHANNEL_NAME = "Core Background Service"
 
     /** Protocols Scheme **/
     const val VMESS = "vmess://"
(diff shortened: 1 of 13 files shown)


eliotcougar:
Compatibility note for #6204: the two branches currently have a content conflict in `NotificationHelper.kt`. This is a small integration issue rather than a behavioral contradiction.

Preferred reconciliation: merge/rebase this channel-localization PR first, then rebase #6204 and have its transient-message channel use `ensureNotificationChannel(...)` with `R.string.notification_channel_other`. #6204 should retain its notification-permission checks, posting, timeout, and cancellation behavior. The locale XML changes merge cleanly.

2dust:
这个完全没有必要，特别是在 NotificationChannelType 还加入了 string 。

eliotcougar:
> 这个完全没有必要，特别是在 NotificationChannelType 还加入了 string 。

I'm not sure I understand what you mean. This PR makes notification channel names localized in the Android notification settings. Right now they are always in English.

This PR follows the Android guidelines for localized notification channel names. If the user changes the OS language, the notification channel should change without resetting the notification channel preferences (sound, vibration, importance, etc)

eliotcougar:
Perhaps I explained the scope poorly. These are the channel names users see in Android Settings → Apps → v2rayNG → Notifications, not internal identifiers or notification message text.

`NotificationChannelType` already contained hard-coded English names. This PR replaces those names with translated string-resource references. Channel IDs remain unchanged.

Would you prefer keeping the resource mapping in `NotificationHelper` instead of the enum?

PR #6200, opened 2026-09-06
fix: Keep proxy-chain deletion targets tied to stable row keys

eliotcougar:
Store the pending member key instead of its list position and resolve that key when deletion is confirmed. Remove the member and its corresponding row key together; if the target no longer exists, leave the draft unchanged. Reuse the same key-based removal for blank members.


eliotcougar:
Compatibility note for #6202: Git can auto-merge these branches, but the resulting Kotlin source does not compile. This PR deliberately replaces `memberToDeleteIndex` with `memberToDeleteKey`; #6202's named chain-member confirmation is then inserted into that block while still referring to the removed `index` variable.

Preferred reconciliation: merge/rebase this stable-key fix first, then rebase #6202 and resolve the displayed member name through `memberToDeleteKey` plus `memberKeys`, never through a captured list index. The combined regression coverage should include reordering while the confirmation is open, verifying that both the displayed name and deletion target still refer to the same stable member.

PR #6202, opened 2026-09-06
UI/UX: Add item names to deletion confirmations

eliotcougar:
Carry display names alongside existing GUIDs in main/subscription list confirmation requests, and retain the main target across recreation with its two-string saver. Use saved names in editors rather than uncommitted renames.
Center delete confirmation messages while retaining a single Text node for the question and target. Visually separate item name with a blank line in all nine locales, including asset files, which previously embedded the filename in the question.
Keep the title-free dialog and existing Cancel/Delete actions, and leave ordinary confirmation messages start-aligned. Bulk-delete prompts retain their collective wording.

This is a feature extraction from the accessibility PR set.


eliotcougar:
Compatibility note for #6200: Git can auto-merge these branches, but the resulting Kotlin source does not compile. #6200 replaces `memberToDeleteIndex` with `memberToDeleteKey`, while this PR's named chain-member confirmation still formats `members.getOrNull(index)` after the automatic merge.

Preferred reconciliation: merge/rebase #6200 first, then rebase this PR and derive the confirmation name from the stable member key through `memberKeys`. The combined test should reorder members while the dialog is open and verify that the shown name and eventual deletion still target the same member. The other named-confirmation call sites are not adversely affected by #6200.

2dust:
改进是好的，但是有点过度了，比如 string 分的太细了，每个语言的翻译和维护变难了

eliotcougar:
> 改进是好的，但是有点过度了，比如 string 分的太细了，每个语言的翻译和维护变难了

I don't think it's such a big problem. It's only a simple string concatenation.

It is an extra chance given to the user to avoid a mistake.

Alternative is to make it two separate elements, but in that case TalkBack will not read it as a single block of text.

eliotcougar:
Moved message construction from the resources to the dialog component. This way it looks cleaner, I guess.

PR #6203, opened 2026-09-06
fix: Keep connection-test requests tied to their lifecycle

eliotcougar:
Correlate current and bulk replies with the request that produced them so older progress, cancellation, and completion cannot finish a replacement test.

Own current probes in the core manager and cancel them on replacement, stop, or reload. Use the existing ordered service acknowledgement to retire requests when the daemon is absent. Cancel pending bulk preparation when the UI cancels or its ViewModel is cleared, and snapshot the filtered target list before asynchronous preparation.

Tag bulk cancellation per worker, including service destruction, instead of reporting cancellation as successful completion or issuing an untagged reply. Report rejected bulk-service starts through the same cancellation path.


eliotcougar:
Compatibility note for #6205: these branches have substantive conflicts in `CoreTestService`, `MainRepository`, `MainServiceEvent`, and `MainViewModel`. This PR adds request IDs and lifecycle/stale-result filtering; #6205 changes the same result contract to carry GUID/delay data through a lossless, coalesced UI path. Neither implementation should simply overwrite the other.

Preferred reconciliation: treat this lifecycle contract as the foundation, then rebase #6205. The combined success event should carry both `RealPingResult` and `requestId`; the repository should retain #6205's lossless channel; the ViewModel should validate the active request before queueing a result, bind pending results to that request's group, force the terminal flush before completion, and clear the matching batch on cancellation/replacement.

PR #6221, opened 2026-09-11
Enable disableFallbackIfMatch for built-in DNS

ItsVixano:
Currenlty, domestic domains are resolved with remote DNS which is not the expected behavior. It is especially bad in situations when proxy provides IPv6 but there is no local IPv6 connectivity. In such cases domestic hosts with AAAA records become inaccessible. disableFallbackIfMatch fixes this by removing remote DNS from domestic domains resolution.

Change-Id: If24ec8f1eacf81c3e25a27820cd98a9f5b878ecd


ItsVixano:
@2dust hi, why did you close this PR? Did you come up with a better solution? :3 ~

Issue #6233, opened 2026-09-14
[Bug] 自定义配置信息缺失与延迟测试失效，URL导入订阅严重卡顿 | [Bug] Missing info in Custom Configs, Delay test failure, and severe lag in URL subscriptions


Astiyaxh:
预期行为 (Expected Behavior)
In the new UI, Custom Configs should display the Address and Port under the config name.

The TCP delay test and Real Delay test should work properly and quickly for Custom Configs.

Adding a new subscription via the URL scheme (v2rayng://install-config?url=...) should be fast and instant, regardless of how many subscriptions are already saved in the app.

实际行为 (Actual Behavior)
Since the new UI update, Custom Configs do not show the server Address and Port. (Note: standard configs show these details perfectly. I have provided a sample standard config below in the "Steps to Reproduce" section so you can compare the UI behavior with a Custom Config).

The TCP delay test does not work for Custom Configs (shows nothing), and the Real Delay test is extremely slow and performs poorly.

When V2RayNG already has 10 subscriptions, adding an 11th via v2rayng://install-config?url=... takes around 50 seconds and causes the app to freeze/lag. However, when there is only one subscription, adding another through the same link is fast and works normally.

复现方法 (Steps to Reproduce)
Add a Custom Config and observe the missing address/port info in the new UI. To see the contrast, you can add this standard config which correctly shows the details

Run TCP Ping and Real Delay tests on your Custom Config.

Keep 10 subscriptions in the app, then click on a v2rayng://install-config?url=... link to add the 11th one, and observe the ~50 seconds freezing/lag. Then clear subscriptions, keep only 1, and try adding another via the same link to see it work instantly.

日志信息 (Log Information)
环境信息 (Environment Information)
v2rayNG version: 2.3.8

Android version: 16

Phone Model: Samsung Galaxy A54 5G

额外信息（可选）(Additional Info)
N/A

<img width="576" height="1280" alt="Image" src="https://github.com/user-attachments/assets/e37adfdd-4987-40b0-afdf-765d3f0df6de" />

eliotcougar:
For stability reasons, CUSTOM profiles cannot be batched during Real Delay tests because they can have conflicting extra options. Real Delay testing a subscription group containing many CUSTOM profiles will take a long time because they have to be tested one by one.

PR #6236, opened 2026-09-15
Enable TCPing for single-server custom profiles

eliotcougar:
## Summary

TCPing currently skips every custom profile. Resolve its target from the saved raw JSON so configs with exactly one valid server address and port can be tested. Multiple-server configs and invalid endpoints remain skipped, and stale cached addresses are never used as a fallback.

**Depends on #6235.** Until that parser fix is merged, this PR also includes the parser changes. The [TCPing-only change](https://github.com/eliotcougar/v2rayNG/commit/986c663814cf8cd4d8b40c88e30a460b03014acb) is limited to the worker and its JVM regression tests.

Related to #6233 (custom-profile TCPing).

## Validation

- 107 JVM tests passed, including 8 new TCPing endpoint cases.
- Play Store debug and release builds passed; native packaging checked for all ABIs.
- Emulator UI tests confirmed successful TCP connections for two single-server custom configs and a normal control, with no connections to multiple-server or stale cached endpoints. Stored results and raw JSON were checked afterward.

**Not run:** physical-device, public-server connectivity, or native real-delay runtime tests. Emulator validation used a local TCP listener with the app's declared local-network permission granted.


2dust:
抱歉，tcping 没有计划支持 自定义。
并且 tcping 属于不推荐的用法，很有可能会被删除

PR #6246, opened 2026-09-19
chore(deps): MMKV 1.3.17, mockito-kotlin 6.3.0, drop mockito-inline

aleksandr-miheichev:
Two unrelated dependency updates, neither changing production behaviour. MMKV goes 1.3.16 -> 1.3.17 for data-safety fixes in the native layer that backs every `MmkvManager` store. On the test side `org.mockito:mockito-inline` is removed as unmaintained and redundant, which also frees `mockito-kotlin` from the version reference it shared with that artifact and moves it 5.2.0 -> 6.3.0.

## MMKV 1.3.16 -> 1.3.17

[v1.3.17](https://github.com/Tencent/MMKV/releases/tag/v1.3.17) carries three fixes: oversized keys are rejected before their length can overflow and corrupt MMKV data; auto-expiration timestamps no longer overflow for very large durations; native instance close lifecycle and recursive thread-lock destruction are hardened. The first and third matter here — every store in `MmkvManager` is opened in `MULTI_PROCESS_MODE`, and corruption reaches the app through the `MMKVHandler` it installs, whose `onMMKVCRCCheckFail` / `onMMKVFileLengthError` answer `MMKVRecoverStrategic.OnErrorRecover`, so a corrupted store costs the user whatever cannot be recovered. The auto-expiration fix is inert for this app: there is no `enableAutoKeyExpire` or `*ExpireIn` call site under `V2rayNG/app/src/`.

Why 1.3.17 and not the current 2.4.2: 1.3.x is MMKV's declared LTS line, and [v1.3.7](https://github.com/Tencent/MMKV/releases/tag/v1.3.7) states that 2.0 and later "will discontinue support for 32-bit architectures and raise the minimum SDK version requirement to 23". The app packages `armeabi-v7a` and `x86` alongside the 64-bit ABIs, so 2.x would mean dropping two of the four ABIs shipped today. 1.3.17 is the newest release on the LTS line.

The surface the app touches is unchanged: the `MMKVHandler` interface and the five-argument `MMKV.initialize(context, rootDir, libLoader, logLevel, handler)` overload called from `MmkvManager.initialize` are identical between the two versions. The artifact still ships exactly `arm64-v8a`, `armeabi-v7a`, `x86` and `x86_64`, and its manifest still declares `minSdkVersion 21` — below the app's `minSdk 24`.

## Test tooling: mockito-inline removed, mockito-kotlin 5.2.0 -> 6.3.0

`org.mockito:mockito-inline` has not been published since 5.2.0 (2023-03-09) and is redundant: [mockito-core 5.0.0](https://github.com/mockito/mockito/releases/tag/v5.0.0) switched the default mock maker to the inline one, which is what the three test classes calling `mockStatic` / `mockConstruction` (`ShadowsocksFmtTest`, `SubscriptionIndexTest`, `SettingsViewModelTest`) already rely on. That default covers JVM local unit tests, which is all this repository has — there is no `androidTest` source set and no `mockito-android`.

Removing it is not only cleanup. `mockito-inline` and `mockito-kotlin` shared one version reference (`mockitoMockitoInline = "5.2.0"`), so `mockito-kotlin` was pinned by an artifact it has nothing to do with. Giving it its own reference is what makes 6.3.0 reachable. Three transitive consequences, all satisfied by the project as it stands: mockito-core moves 5.7.0 -> 5.23.0 and from `runtime` to `compile` scope, so it is now on the test compile classpath rather than only the runtime one; `kotlin-reflect:2.1.20` is pulled in at runtime; and 6.3.0 ships Kotlin 2.1 metadata and Java 11 bytecode, against the project's Kotlin 2.4.20 and `JvmTarget.JVM_17`. No module other than `:app` referenced the removed alias.

## Verification

Clean checkout prepared the way `.github/workflows/build.yml` does it: submodules initialised recursively, `libv2ray.aar` from the `AndroidLibXrayLite` gitlink `d0c6c4ae1b09c912070c8288bd0dbcc2e492ac29` (tag `v26.9.9`), JDK 21, `platforms;android-37.0`, `build-tools;37.0.0`. Gradle run from `V2rayNG/` on branch HEAD `332fd2b`.

- `:app:testPlaystoreDebugUnitTest` — 81 tests, 0 failures, 0 errors, 0 skipped. This is the check that exercises the Mockito change: the three `mockStatic` / `mockConstruction` classes pass on mockito-core 5.23.0 without `mockito-inline` on the classpath.
- `:app:assemblePlaystoreDebug` — pass; four per-ABI APKs plus the universal APK.
- Native packaging, per ABI: `libmmkv.so` in the built APKs is byte-identical (SHA-256) to the one in the 1.3.17 artifact on all four ABIs, and matches 1.3.16 on none. Hashes below, so the check is repeatable against the published AAR.

**Not run:** any scenario that executes the new `libmmkv.so`. The repository has no `androidTest` source set, and the unit tests replace MMKV with a static mock (`mockStatic(MMKV::class.java)` in `SubscriptionIndexTest`), so nothing in the suite loads the native library — and an assembled APK does not verify runtime behaviour. Storage read/write, multi-process access and the recovery handler on 1.3.17 are unverified here.

**Not run:** `:app:testFdroidDebugUnitTest` and `:app:assembleFdroidDebug` on this HEAD. Neither change is flavour-scoped — MMKV is declared once for all flavours and the Mockito change is test-only — so the F-Droid variant is not expected to differ.

<details><summary>libmmkv.so SHA-256 in the built APKs</summary>

```text
arm64-v8a    4943cc7195c24431d1569a0659b45fc49a687ac58089515a8b5c57f672ca1dba
armeabi-v7a  64aa9ab475ebde3595120e18d1f3576f6a466cc7481a0e126d52e679b8e2f238
x86          68f9e6d13a3fe1b2a52de058d367421773b3c01e058849dcb228a7ccba10395d
x86_64       4b639aa8fd190b7a6f28dc88be7668e23f26da2edc0d8976cc24dd342089a310
```
</details>


aleksandr-miheichev:
@2dust Thanks for merging #6248. Could you tell me briefly why this PR and #6247 were closed? I'd like to know what is worth sending you in the future. Which is closest?

Would MMKV 1.3.17 alone, or the test-only Mockito cleanup alone, be OK as a separate PR?

